PR #9, opened 2025-11-03
Add arm support

jondkinney:
Needed to update some of the scripts to be able to resolve split package deps


ryanrhughes:
Closing: this predates the package manifest scheme (.omarchy/package.json) and the aarch64 support that has since landed on master for many of these packages. 27 of its 32 packages have no manifest. Thank you for the early ARM work; individual packages from it that are still wanted should come as separate PRs against current master.

PR #41, opened 2026-01-31
Fix invalid placeholder in chromium.desktop

litescript:
The omarchy-chromium binary tarball currently ships chromium.desktop with an unresolved placeholder line (@@EXTRA_DESKTOP_ENTRIES@@), which makes the desktop entry invalid and can trigger parsing errors in consumers like xdg-desktop-portal-gtk. This strips the placeholder during packaging so the installed desktop file is always valid.


litescript:
Root cause seems to be in the release artifact from omarchy-chromium (the .desktop inside the tarball contains the unresolved placeholder). This PR defensively strips it during packaging so installs are valid even before the upstream release pipeline is corrected.

PR #84, opened 2026-04-30
Fix: gpu-screen-recorder is breaking hibernation on some NVidia systems

NicolasDorier:
I found out the root cause preventing my system from hibernating and documented it on https://github.com/basecamp/omarchy/discussions/5500.

This commit hopefully avoid `gsr-nvidia.conf` from being installed with gpu-screen-recorder. ([ref commit](https://git.dec05eba.com/gpu-screen-recorder/commit/?id=5e3a3714dd020d1a1b45f8cbc6239ba15da580db))


NicolasDorier:
Need also this merged https://github.com/basecamp/omarchy/pull/5519, without it, the default value become 2 and this unfortunately same effect as 1 set by this file.

PR #85, opened 2026-04-30
Migrating to a fork for `asdcontrol`

Kasui92:
Since [the original repository has been archived](https://github.com/nikosdion/asdcontrol) and [won’t receive further updates](https://github.com/nikosdion/asdcontrol/commit/0ee8bd576d4e93513027d713e688988cb0d827ef) (for now), I collected some pending fixes along with [this contribution](https://github.com/omakasui/asdcontrol/pull/1) and moved them into a [fork](https://github.com/omakasui/asdcontrol), including a proper release to avoid pointing directly to commits.

This should also make things easier to manage.

### DIFF
--- a/pkgbuilds/edge/asdcontrol/.SRCINFO
+++ b/pkgbuilds/edge/asdcontrol/.SRCINFO
@@ -1,8 +1,8 @@
 pkgbase = asdcontrol
 	pkgdesc = Control brightness on Apple Displays connected via USB-C
-	pkgver = 0.23.0ee8bd5
+	pkgver = 0.6.0
 	pkgrel = 1
-	url = https://github.com/nikosdion/asdcontrol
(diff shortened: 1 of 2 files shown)


pkgbuilds/edge/asdcontrol/.SRCINFO
```diff
@@ -1,16 +1,16 @@
 pkgbase = asdcontrol
 	pkgdesc = Control brightness on Apple Displays connected via USB-C
-	pkgver = 0.23.0ee8bd5
+	pkgver = 0.6.0
```
dhh, line 4: This would mean the release is older than what we already have, and I don't think that'll work? 0.23 > 0.6.

Kasui92 (commented):
@dhh
I added epoch so it looks newer than the legacy version

PR #90, opened 2026-05-06
linux-ptl: add SOF backports for Zenbook S14 UX5406AA audio

Pegorim:
## Summary

Adds two SOF SoundWire patches to `linux-ptl` to fix audio on the ASUS Zenbook S14 UX5406AA (Panther Lake / Core Ultra 9 386H, SSID 10431464). Following up on @dhh's comment in basecamp/omarchy#5557 — happy for you to take this as-is, rework, or use as a reference.

- **0030** = `thesofproject/linux#5728` (still open upstream): tightens SoundWire mask matching so ACE 2.0+ platforms can't fall back to a partial machine table entry.
- **0031** = `thesofproject/linux#5754` (still open upstream): adds the UX5406AA-specific entry. The board's BIOS declares an `rt722` SDCA device on link 0 that doesn't physically exist; the new entry declares it as a zero-endpoint ghost so the SDCA enumeration matches reality (cs42l43 on link 3 + 4× cs35l56 on links 1/2).

Both apply cleanly against the current `linux-7.0.3.arch1` base — only line-offset shifts, no rejected hunks.

## Test results (UX5406AA / BIOS UX5406AA.303)

I built locally against `v7.1-rc2` with the same two patches and audio came up: `sof-soundwire` card registered, all 4 cs35l56 amps loaded their tuning blobs from `linux-firmware-cirrus 20260410-1`, headphone jack and internal mic functional, HDMI works. Full report at thesofproject/linux#5721 (comment 4379439293).

## Caveat — upstream churn

PR #5754 is still under review. @bardliao mentioned in the upstream thread that the next revision (commit `e411f65`) retargets a different topology file (`…-agg-l3-cs35l56-l12.tplg`) which isn't yet in `firmware-sof 2025.12.2-1`. If the merged version of #5754 differs materially from the diff included here, this PR will need a refresh. The version included now is the one I've personally booted and verified end-to-end.

## Refs

- thesofproject/linux#5728
- thesofproject/linux#5754
- thesofproject/linux#5721
- basecamp/omarchy#5557

## Test plan

- [ ] `makepkg -s` against this branch builds successfully on `x86_64`
- [ ] Resulting `linux-ptl` package boots on Zenbook S14 UX5406AA
- [ ] `aplay -l` lists `sof-soundwire` card; speakers/jack/mic produce audio

### DIFF
--- a/pkgbuilds/edge/linux-ptl/0030-sof-intel-exact-sdw-mask.patch
+++ b/pkgbuilds/edge/linux-ptl/0030-sof-intel-exact-sdw-mask.patch
@@ -1,0 +1,37 @@
+diff --git a/sound/soc/sof/intel/hda.c b/sound/soc/sof/intel/hda.c
+index b3d61d973ce40b..24604e71386d6d 100644
+--- a/sound/soc/sof/intel/hda.c
++++ b/sound/soc/sof/intel/hda.c
+@@ -1337,6 +1337,8 @@ static struct snd_soc_acpi_mach *hda_sdw_machine_select(struct snd_sof_dev *sdev
+ 		return NULL;
+ 	}
+ 
++	chip = get_chip_info(sdev->pdata);
++
+ 	/*
+ 	 * Select SoundWire machine driver if needed using the
+ 	 * alternate tables. This case deals with SoundWire-only
+@@ -1352,8 +1354,13 @@ static struct snd_soc_acpi_mach *hda_sdw_machine_select(struct snd_sof_dev *sdev
+ 		 * first check whether link_mask of mach is subset of
+ 		 * link_mask supported by hw and then go on searching
+ 		 * link_adr
++		 * For those platforms that can use the created SDW mach below,
++		 * match the two masks exactly to avoid using an unexpected
++		 * mach.
+ 		 */
+-		if (~hdev->info.link_mask & mach->link_mask)
++		if (~hdev->info.link_mask & mach->link_mask ||
++		    (hdev->info.link_mask != mach->link_mask &&
++		     chip->hw_ip_version >= SOF_INTEL_ACE_2_0))
+ 			continue;
+ 
+ 		/* No need to match adr if there is no links defined */
+@@ -1393,8 +1400,6 @@ static struct snd_soc_acpi_mach *hda_sdw_machine_select(struct snd_sof_dev *sdev
+ 			 peripherals->array[i]->id.part_id,
+ 			 peripherals->array[i]->id.sdw_version);
+ 
+-	chip = get_chip_info(sdev->pdata);
+-
+ 	/* SDCA was not well supported in the BIOS before ACE2.0 */
+ 	if (chip->hw_ip_version < SOF_INTEL_ACE_2_0)
+ 		return NULL;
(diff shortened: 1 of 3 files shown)


dhh:
I see there are open questions on the 5754 PR. Should we still proceed with this? Working on getting Omarchy 3.8 ready. Would be nice to have this in there.

Pegorim:
@dhh the SOF team still figuring out a few details for the final solution. audio is working on my system right now without any trouble. There are bugs here and there, but I can live with it until they find a final solution.

I am using the OS with the patch as my daily driver, so there is no other problems happening because of it. 

If you think it is best to wait for their final solution its fine, but the system works here for me as is.

dhh:
I think it's more important to get the Zenbook S14 working with audio. We can always build a new kernel when there's a better fix. What are the bugs you're seeing?

Pegorim:
bug: I had  my audio through the headphone connected to a beringher XR18 board. I watching a church service through the church sound system (I wanted to test in case we use my machine for streaming)

I had to pan everything to the left so that I had all my audio through only one channel (the berhinger aux entry is balanced so they dont work properly with stereo sound)

For some reason, after listening for a while, the pan that was 100% to the left came back to the center again. 
Does it make sense?

TLDR; I had my audio set up all the way to the left, and it came back to the center without any human intervention.

itisyb:
I have similar issues with Lenovo Slim 7i ultra x9 388, soundacard not being able to be detected 

Zenbook has CS35L56 amps as SoundWire
devices on links 1/2. This Lenovo uses CS35L57/CS35L56 over SPI sidecar

itisyb:
Cross-linking a related Lenovo Panther Lake SOF/SoundWire case: #97

#97 uses the same general exact-mask + ghost RT722/ALC722 idea from this PR, but for Lenovo Yoga Slim 7 Ultra 14IPH11 (`83QK`, SSID `17aa:3821`). The Lenovo layout is different: CS42L43 and the ghost RT722 both show up on SoundWire link 3, while the speaker amps are SPI CS35L57/CS35L56 sidecar devices. It also needed a CS42L43 endpoint-filter adjustment and a `17aa:3821` PCH-DMIC override.

Leaving the link here so other Panther Lake audio users/maintainers reading #90 can find the Lenovo follow-up.

dhh:
Would be great to have a single patch/setup that works for both. But even better if we can get this into upstream, so we don't have to maintain a fork for long!

dhh:
Maybe @TsaiGaggery can help guide us here. What's already coming upstream, what can safely be backported. Would be great to have all Panther Lake laptops just work out of the box!

bardliao:
> Cross-linking a related Lenovo Panther Lake SOF/SoundWire case: #97
> 
> #97 uses the same general exact-mask + ghost RT722/ALC722 idea from this PR, but for Lenovo Yoga Slim 7 Ultra 14IPH11 (`83QK`, SSID `17aa:3821`). The Lenovo layout is different: CS42L43 and the ghost RT722 both show up on SoundWire link 3, while the speaker amps are SPI CS35L57/CS35L56 sidecar devices. It also needed a CS42L43 endpoint-filter adjustment and a `17aa:3821` PCH-DMIC override.
> 
> Leaving the link here so other Panther Lake audio users/maintainers reading #90 can find the Lenovo follow-up.

@charleskeepax FYI

charleskeepax:
Would you guys be able to try this fix:

https://github.com/thesofproject/linux/pull/5773/commits

I think it might be a slightly better approach.

Pegorim:
This approach seems much better @charleskeepax I am testing it, update when ready. 

@dhh this approach will simplify fixing other laptops with the same issue, we will only have to add the board to the dmi-quirks file.

charleskeepax:
@Pegorim Apologies seems I missed a terminator in the first version of the patch there is an updated one now. Also I did my best to check the DMI stuff from the info on the tickets, but there is slightly that might need slightly tweaked if the match doesn't take.

Pegorim:
No problem sir. I am running a few checks before the test.

ryanrhughes:
Superseded: this targets the old pkgbuilds/edge/ layout, which was removed when channels moved to metadata. linux-ptl lives at pkgbuilds/linux-ptl now; if the SOF backports are still needed, please open a fresh PR against current master.

PR #91, opened 2026-05-07
We don't need srcinfo cluttering things up

dhh:
Keep the courtyard clean.


ryanrhughes:
Refactored how the packages are structured which took this into account along the way as a part of #94.

PR #92, opened 2026-05-07
Skip packages where we have upped the pkgrel to something higher

dhh:
Avoids the problem we have with the v4l2-relayd package that sync-aur keeps wanting to update.


ryanrhughes:
Refactored how the packages are structured which took this into account along the way as a part of #94.

PR #98, opened 2026-05-12
bump pkg and packages to 7.0.5

spencerbull:
## Summary
- Bumps `linux-ptl` from `7.0.3.arch1-2` to `7.0.5.arch1-1`.
- Refreshes the kernel config and Panther Lake patch context for the Arch `7.0.5` kernel.
- Aligns `linux-ptl` build and headers dependencies with Arch’s `linux` PKGBUILD, excluding docs-only deps we do not build.

## Notes
The expanded `makedepends` and `linux-ptl-headers` dependencies are not unique to `7.0.5`; Arch already declares them for its kernel package. This avoids relying on packages that happen to exist in the build image and makes the Rust, BTF, compression, crypto, and external module tooling dependencies explicit.

## Tested Platforms
- XPS 16 OLED, XPS 14 OLED
- XPS 14 IPS


spencerbull:
removing and bumping to 7.0.9

PR #118, opened 2026-07-21
Add omarchy-emacs

scottjones:
## Summary

Adds **omarchy-emacs** as an AUR-synced package (`source: aur`), per @dhh's request in basecamp/omarchy#6329 — so Emacs-with-Omarchy-integration installs from the Omarchy repo via the regular installer instead of relying on the AUR at install time.

## What it is

`omarchy-emacs` auto-syncs Emacs to Omarchy's active theme and font (Wayland-native `emacs-wayland`/pgtk). It ships an `omarchy.el`, a theme file, and theme/font hooks; works on both Omarchy 3 and 4.

- AUR: https://aur.archlinux.org/packages/omarchy-emacs
- Source: https://github.com/scottjones/omarchy-emacs

## Package

- Created with `bin/add-package omarchy-emacs` — `.omarchy/package.json` is `{ "source": "aur", "upstream_commit": "ec60427…" }`.
- The synced PKGBUILD builds `arch=any` (all Emacs Lisp + shell, no compiled binaries); depends on `emacs-wayland` and `bash`.
- Left in the default release ring (edge → promote). Happy to add it to the `fast` ring if you'd prefer it to reach stable directly.

## Test plan

- [x] `bin/add-package omarchy-emacs` synced cleanly from AUR
- [x] `makepkg` builds; source sha256 verified (`Passed`)
- [x] Verified end-to-end on Omarchy 4 — theme and font tracking both work

## Follow-up

Once this lands, I'll update basecamp/omarchy#6329 to install via the regular `omarchy-pkg-add` path instead of `omarchy-pkg-aur-add`, so the menu entry no longer touches the AUR.


pkgbuilds/omarchy-emacs/omarchy-emacs.install
```diff
@@ -0,0 +1,23 @@
...
+    a shim that loads the package-installed code, so the new code is
+    already on disk; only in-memory state needs to refresh.
+
+    Omarchy 4 note: 1.8.7 tracks the theme at its new location
+    (~/.local/state/omarchy/current/). If your Emacs shows "Omarchy colors
+    not available" after a theme change, your ~/.config/emacs/omarchy.el is
+    a stale pre-shim copy — run 'omarchy-emacs-setup' once to convert it to
+    the shim, then 'omarchy-restart-emacs'. This is also required for anyone
+    upgrading from < 1.8.5.
+EOF
```
dhh: Hmm, don't love all of this. Can we not just bake whatever we need to have done into the installer that's inside of Omarchy?

scottjones: Good idea — dropped it. Everything the scriptlet nagged about is already baked into the Omarchy installer: `omarchy-install-emacs` runs `omarchy-emacs-setup`, which writes the user-side shim, seeds the config, and installs the theme/font hooks. The pacman `post_install`/`post_upgrade` messages were just duplicating that with a nag (plus a legacy note for pre-shim upgraders that's no longer relevant).

Re-synced to 1.8.9, which removes the `.install` entirely — the package is now just `PKGBUILD` + `.omarchy/package.json`, matching the omarchy-nvim shape.

scottjones:
I left this on the default ring (edge → `bin/repo migrate` to stable). Happy to set `"release_ring": "fast"` if you'd rather it hit stable directly — that'd also let basecamp/omarchy#6329 switch to the regular `omarchy-pkg-add` installer sooner.

PR #71, opened 2026-04-01
Skip transparency overrides in Neovide

mplaczek99:
Skip transparency overrides in Neovide

Prevent transparency highlight overrides from running in Neovide.

This avoids the black background issue caused by clearing bg on groups like
Normal and NormalNC, while keeping existing terminal Neovim behavior unchanged.

Closes basecamp/omarchy#5107

@dhh I’ve moved the PR and adjusted it accordingly. Would appreciate a review when you have a moment.

PR #124, opened 2026-08-07
Add HP EliteBook X G2i audio and camera support

meirdick:
Adds support for the HP EliteBook X G2i, a Panther Lake laptop. On a stock
install it has no working speakers, no microphone and no webcam.

Companion PR: basecamp/omarchy#6596.

## What this adds

Three kernel patches to `linux-ptl`, in the same style as #116, #90 and #97:

| Patch | What it does |
|---|---|
| `0030` | Adds this board to the Panther Lake ACPI match table. It pairs an RT712 codec with four TAS2783 amps, a combination the table does not list, so SOF never starts the amps and the card has no Speaker device at all. |
| `0031` | One line. Gives the TAS2783A amp a component name, so the card reports `spk:tas2783` and ALSA can find a speaker configuration. Every other amp in the list already has one. |
| `0032` | Intel's own patch, carried unchanged. Adds the second link frequency the OV05C10 camera needs. Without it the sensor fails to probe. |

`linux-ptl` tracks 7.1.8.arch1, the version this repo now publishes, and takes
upstream's PKGBUILD and config verbatim. Only the three patches above are added,
so `pkgrel` is 2.

`intel-ipu7-drivers` is new: the DKMS modules split out of `intel-ipu7-camera`.
The IPU7 has no in-tree driver, so anything wanting frames needs these modules,
but until now they only came bundled with CamHAL, icamerasrc and v4l2-relayd.
This machine needs the modules without CamHAL. Same pinned commits and same
`/usr/src` paths, so nothing rebuilds for existing users. `intel-ipu7-camera`
now depends on it.

`libcamera-ov05c10` is new: libcamera 0.7.2 with five patches, built into
`/usr/lib/libcamera-ov05c10` rather than over the top of Arch's libcamera. It
neither provides nor conflicts with it, and nothing that does not deliberately
point at it can be affected. The patches add an OV05C10 sensor helper and lens
shading correction, neither of which libcamera has.

`hp-elitebook-x-g2i-audio` is the Speaker device definition, firmware filename
symlinks, and two guards against a boot race where the amps enumerate after
WirePlumber starts. No DKMS. Nothing here overwrites a file another package owns.

`hp-elitebook-x-g2i-camera` is Intel's `ov05c10` sensor driver as DKMS, plus a
daemon that drives the sensor through libcamera and publishes the result on a
v4l2loopback device. Colour comes from HP's own factory calibration, decoded
from the tuning file HP ships on the machine's Windows partition.

## Testing

Built, installed and cold-booted on the reference machine, SKU `D4DU1UT#ABA`,
on 7.1.5.arch1-2. Speakers, microphone and webcam all work. Audio needs no DKMS module: the
kernel's own match table plus one new UCM file. The camera is listed and working
in Chrome, Firefox and OBS, and costs about 0.33 of one CPU core while
streaming. The privacy LED is dark until an application opens the camera.

Every file both stacks depend on is owned by a package in this PR.

Following #116's checklist: kernel tarball and Arch patch verified with SHA-256,
BLAKE2 and PGP; all nine patches apply without rejects; the tracked config
matches Arch's; `makepkg` produces `linux-ptl 7.1.8.arch1-2`.

**On 7.1.8 specifically:** this branch was rebased onto it after upstream bumped
the package mid-review. All three patches apply with no rejects, and the result
has now been built, installed and cold-booted on the reference machine:

```
kernel   7.1.8-arch1-2-ptl
audio    cfg-amp:4 spk:tas2783, HiFi profile, Speaker sink, tuning loaded,
         Mic and Headset sources present
camera   /dev/video50 "HP EliteBook X G2i Webcam", 30 frames captured,
         privacy LED dark at idle, lit while streaming, dark again after
DKMS     ipu7-drivers, ov05c10, v4l2loopback, vision-drivers all built
units    no failed system or user units belonging to these packages
```

Worth stating plainly, because it is the case this PR exists to prevent: when
`omarchy update` pulled upstream's `linux-ptl 7.1.8.arch1-1` on this machine, the
ACPI match module contained no `ptl_sdw_rt712_l3_tas2783_l2` entry, and booting it
would have left the machine with no Speaker device at all and a camera sensor that
fails to probe. Stock Arch 7.1.8 was used as a control and lacks it too. Every
kernel bump does this until the patches land here.


meirdick:
Marking draft: the PKGBUILDs currently source the DKMS trees from a private repo, so `makepkg` fails at the clone step and the packages do not build. Splitting the GPL-2.0 kernel sources into a public repo and will flip back to ready once `makepkg` passes.

meirdick:
Fixed in ace6262. The DKMS trees are GPL-2.0 throughout, so they are split into [meirdick/hp-elitebook-x-g2i-dkms](https://github.com/meirdick/hp-elitebook-x-g2i-dkms) and fetched at a pinned commit.

Both packages now build:

```
==> Finished making: hp-elitebook-x-g2i-audio 1.0.0-1
==> Finished making: hp-elitebook-x-g2i-camera 1.0.0-1
```

Payloads verified: DKMS trees land in `usr/src/<name>-1.0/` matching each `dkms.conf` `PACKAGE_NAME`/`PACKAGE_VERSION`, and anonymous clone of the source repo is confirmed working.

meirdick:
Back to draft. The audio package currently rebuilds the whole `snd-soc-acpi-intel-match` module from vendored kernel sources and installs it to `/updates`, which overrides the match tables for every Intel platform (adl through nvl), not just this board. That freezes those tables at the forked kernel version for anyone who installs it. Reworking before review.

meirdick:
Background on a few things in this PR that will otherwise raise questions.

**Why the camera goes through userspace at all.** CamHAL produces black frames on
this machine. The cause is not established. An earlier version of the camera
README claimed the graph settings expect a 2944-wide readout while the sensor
gives 2888, producing a negative crop. That explanation is wrong: the four logged
numbers are edge insets, not a rectangle, so a right inset of -56 means "read 56
columns past the sensor width", and 2888 + 56 = 2944. Both Intel's and HP's graph
binaries declare one sensor geometry, 2888x1808, and every Intel sensor uses the
same idiom, including OV08X40, which works. Tested August 2026 against a graph
binary identical to current upstream, not retested since. Intel issues #48, #52,
#70 and #71 all cover IPU7 black frames and are open and unanswered.

**Why `intel-ipu7-drivers` declares a conflict.** The AUR `intel-ipu7-dkms-git`
and this tree both install `intel-ipu7.ko`, and DKMS lets the last one win with
no warning. A kernel upgrade reinstalled them in the opposite order and the
camera went black: the AUR build is compiled without `CONFIG_INTEL_IPU_ACPI`,
which disables the code that binds the sensor, and nothing logs an error. With
the conflict declared, pacman refuses the combination instead.

**What the camera costs.** About 4.4 CPU cores at 1920x1080. That is inherent to
doing the ISP in software; reordering the filter chain measured marginally
slower, not faster. Colour is uncorrected: static white balance gains, no colour
matrix, no gamma. HP's factory tuning does contain usable colour matrices, but a
first attempt at applying them made the image visibly worse, so none of it is
here. libcamera looks like the better long-term answer and is being looked at
separately.

**Why patch `0031` matters more than its size suggests.** The audio package used
to be uninstallable. It shipped a modified copy of `sof-soundwire.conf`, which
`alsa-ucm-conf` owns, so pacman refused the transaction. With the card reporting
`spk:tas2783`, stock alsa-ucm-conf resolves a Speaker device from the one new
file the package adds, and the copy is gone.

**Upstream status.** Patches `0030` and `0031` are mine to send to alsa-devel and
are not filed yet. `0032` is Intel's. Nothing here depends on that; each one
accepted later removes a layer from this packaging.

Sensor bind on the reference machine, for reference:

```
intel_ipu7_isys: bind ov05c10 1-0010 nlanes is 2 port is 0
intel_ipu7_isys: All sensor registration completed.
```

meirdick:
Validation update. Everything below was run on the reference machine today, on
7.1.5-arch1-2-ptl built from this branch.

**The audio package is now installed as a package.** It had only ever been
proven as hand-placed files, which is not the same claim. `pacman -U` installs
it with no file conflict against any package, and the Speaker sink survives the
WirePlumber restart the install triggers.

**Patch 0031 does what it says.** The card now reports:

```
components: HDA:80862822,80860101,00100000  cfg-amp:4 iec61937-pcm:7,6,5
            hs:rt712 mic:rt712 spk:tas2783
```

`spk:tas2783` is the whole point of the one-line patch. It is what makes stock
`alsa-ucm-conf` include this package's `tas2783.conf`, which is why nothing here
overwrites a file `alsa-ucm-conf` owns.

**The old DKMS module is gone.** `snd-soc-acpi-intel-match-hp-g2i` was still
registered from the previous approach; it has been removed and the in-tree
`snd-soc-acpi-intel-match.ko` restored. `dkms status` now lists only
`ov05c10`, plus `ipu7-drivers`/`vision-drivers` from `intel-ipu7-drivers` and
`v4l2loopback`. No package here overrides an in-tree module.

**The camera README now records a route off the software ISP.** Saying CamHAL is
an open problem and stopping there is not useful to whoever picks this up.
libcamera's `simple` pipeline handler already supports `intel-ipu7` and drives
this sensor end to end at 0.8 CPU cores against this package's 4.4, debayering
on the GPU. What is missing is a `CameraSensorHelper` for `ov05c10`, which
belongs upstream in libcamera rather than in this repo, and the idle handling
that is most of the code in this package. Both are written down, along with the
measured black level and the native-mode constraint, so the next attempt starts
from data rather than from scratch.

Package build and install is clean for all four: `linux-ptl 7.1.5.arch1-2`,
`intel-ipu7-drivers 1.0.5-1`, `hp-elitebook-x-g2i-audio 1.1.0-1`,
`hp-elitebook-x-g2i-camera 1.1.0-1`.

meirdick:
Found and fixed a fresh-install blocker while auditing this for merge. Worth
recording because of how it hid.

**The camera package shipped options for `v4l2loopback` but nothing that loads
it.** `v4l2loopback-dkms` builds the module and installs no `modules-load.d`
entry. On the reference machine the module was loaded by two things the package
does not own: a hand-made unit left over from the pre-packaging work, and
`/usr/lib/modules-load.d/v4l2-relayd.conf`, which arrives with
`intel-ipu7-camera`. A fresh EliteBook has neither, so `/dev/video50` would
never have appeared and every camera application would have found no device.

This is the same shape as the dependency gap noted when this rework started:
invisible on a machine that already carries the packages from an earlier
approach. It only surfaced after removing every hand-placed file on this machine
and checking what was left.

Fixed with `ExecStartPre=/usr/bin/modprobe v4l2loopback` in the unit rather than
a `modules-load.d` entry, since this service is the module's only consumer and
loading on start also covers a kernel upgrade that rebuilt the module after
boot. Also added `softdep ov05c10 pre: intel_ipu7`, which had likewise existed
only as a hand-made file here.

Proven rather than reasoned: with the service stopped, `modprobe -r
v4l2loopback` removed the module and `/dev/video50` with it; starting the
service brought both back and captured frames.

`pkgrel` bumped to 2.

**Also cleaned off this machine, so it now represents a fresh install:** four
hand-made `/etc/modprobe.d` drop-ins, two stale `/usr/lib/modprobe.d` files, the
`camera-init.service` unit, five `/usr/local/bin/hp-g2i-*` scripts from the
pre-packaging era, and a leftover UCM backup file. Every file the audio and
camera stacks now depend on is owned by a package in this PR.

meirdick:
Background on the camera, since the approach changed during this PR.

**It used to run ffmpeg and cost about 4.4 CPU cores.** Every ISP stage ran in
software at sensor resolution. It worked, and the cost was inherent to doing the
work on the CPU.

libcamera's `simple` pipeline handler already lists `intel-ipu7` and drives this
sensor end to end, debayering on the GPU. The same 1920x1080 output now costs
about 0.33 of one core. What libcamera was missing is why `libcamera-ov05c10`
exists:

| Patch | Why |
|---|---|
| `0001` | No sensor helper for ov05c10, so AGC read the gain range as 16-248 instead of 1-15.5 and settled on the wrong curve. |
| `0002` | The simple IPA has no lens shading algorithm at all, while HP's tables call for up to 8.65x at the corner. |
| `0003` | The GPU path computes statistics from the raw input, so without this the shader's correction never reaches auto white balance. |
| `0004` | `ov05c10` reports an empty analog crop with a success return, which defeats libcamera's fallback and leaves `PixelArrayActiveAreas` as `0x0`. The mapping was correct here only because a guard caught it; now it says so. |
| `0005` | Separates colour shading from luminance shading, which the next paragraph needed. |

**A private prefix rather than replacing libcamera.** `CameraSensorHelper` lives
in `libipa`, a static library linked into each IPA module, so in principle only
`ipa_soft_simple.so` needs to change. In practice libcamera verifies IPA modules
against a public key compiled into `libcamera.so`, and Arch discards the private
half after building, so a separately built IPA loads out-of-process. A private
prefix builds its own keypair and keeps the IPA in-process, without shadowing a
package `pipewire-libcamera` links against.

**The shading tables ship inert, and that is deliberate.** A flat field measured
on this machine showed the colour half made things 1.93x worse under LED light.
HP calibrated tungsten, daylight and five fluorescents, but no LED. Sorting HP's
own tables shows the effect tracks source spectrum, not colour temperature —
continuous sources carry 11-17%, fluorescents 3.5-8.5% — while the algorithm
selects a table by colour temperature. The luminance half is validated and is a
noise-versus-flatness policy call, so it is left at 0 with the cost documented
beside the knob.

**A PipeWire/portal delivery route was tested and rejected.** It works: one clean
node, correct privacy LED, no daemon at all. It was rejected because it needs a
browser flag per browser and reaches only PipeWire-aware applications, whereas
the v4l2loopback device works in everything with nothing to configure.

**Two coupled requirements, if you are reviewing the daemon.** `exclusive_caps=1`
is required for Chrome to enumerate the device at all, and it reports
capture-only only while a writer is attached, so the daemon holds the loopback
open permanently. And the privacy LED follows the sensor, so the sensor is
released when nothing is reading — which means the source behind the loopback
switches between black frames and the sensor, and both directions stop on a
frame boundary or the stream desyncs permanently.

**Earlier revisions of this PR carried a claim that CamHAL cannot work here,
citing a negative crop in the graph settings.** That was wrong and is retracted.
The four numbers in that log line are edge insets, not a rectangle, every Intel
sensor uses the same idiom including ones that work, and the HAL copies the value
verbatim. CamHAL producing black frames on this machine is an open problem, not a
closed one.

meirdick:
This branch grew since the last push. Everything below is verified on the machine.

**Audio now survives suspend.** Before: wake took ~9 seconds and the speakers could come back silent. The kernel package now carries three layers of fixes: one local patch so resume does not wait on the amplifiers (0033), six backports that are already in mainline 7.2/7.3 (0034-0039), and one patch still in review upstream (0040) that brings all four amplifiers back on the first stream after wake. 0034-0039 can be dropped as stable kernels catch up, and 0040 when it merges.

**The camera now uses Intel's hardware ISP.** hp-elitebook-x-g2i-camera 2.0.0 behaves exactly like 1.x from the outside: one writer on /dev/video50, sensor and LED off when idle. Inside, frames come from CamHAL via a new pinned package, hp-elitebook-x-g2i-camhal-runtime. The pin (release 20260327_1) is deliberate: newer releases produce black frames on this sensor (intel/ipu7-camera-hal#48). If the engine fails twice, the old software path takes over for the rest of the boot. The gains: correct colour in any lighting, half the noise, a third of the CPU.

**One small HAL patch rides in the runtime.** The HAL ships with noise reduction set to zero and no way to change it. The patch adds an override; the package defaults to the HAL's own strongest preset after measuring 45% less dark-scene noise and no loss of sharpness. Users can also drop their own AIQ tuning file in /etc; the package ships nothing that cannot be redistributed.

TESTPLAN-v2.md in the camera pkgbuild has the test drills, including rollback. The companion PR basecamp/omarchy#6596 fixes both things Copilot flagged: the ISO mirror package list, and removing a wrongly installed intel-ipu7-camera so setup reruns stop failing on the conflict.

cyberKrakn:
Happy to test this on a second board. My reference here is `D4VJ0UA#ABA` (HP EliteBook X Flip G2i, board `8E86` — same board as your reference), currently on stock Omarchy with the broken `intel-ipu7-camera 1.0.5`:

- camera sensor is `OVTI05C1` (OV05C10) and never binds (no `ov05c10` driver; the package ships only OV08X40 config/tuning), `/dev/video50` green;
- speakers are dead for the same missing ACPI-match reason as your reference machine.

I can build/install the branch, cold-boot, and report back with `dmesg`/unit status, the DKMS build result, and camera in Chrome/Firefox/OBS plus audio (Speaker sink, Mic/Headset sources). Anything specific you'd want captured for a second SKU?

meirdick:
@cyberKrakn thats great, i have a new build off the 7.2 kernel that i'll push later today.

meirdick:
@cyberKrakn pushed. Branch `hp-elitebook-x-g2i` at `038c640`, now on `linux-ptl 7.2.3.arch1-4` (Omarchy's current base). What changed since the Aug 31 push, all verified on the reference `D4DU1UT#ABA`:

**Audio**
- True 4-speaker stereo. Before, all four amps played the left channel. Three kernel patches (`0061`–`0063`) split the amp link across two PDIs.
- Factory calibration now applied. The TAS2783 driver looked for it under Lenovo's UEFI GUID; HP uses `02f9af02-…`. Patch `0065` tries the DELL/HP/LENOVO table. Without it one amp (uid `0x9`, right woofer) is silent.

**Camera**
- Cold-start fix. The ov05c10 needs 50 ms after the INT3472 handshake pin, kernel gave it 25 ms — CamHAL got no frame on first start. Patch `0066`.
- The camera package is otherwise unchanged from Aug 31.

**To capture on the Flip**, in this order:

1. `journalctl -k -b | grep -E 'sdca-.amp-id2'` → want `sof-sdca-2amp-id2.tplg`, not `1amp`
2. `journalctl -k -b | grep -c 'calib data update done'` → want `4`
3. `ls /sys/bus/soundwire/devices/` → does the Flip enumerate `sdw:0:2:0102:0000:01:{9,a,c,d}` like the reference? The match table hardcodes that order (L,R,L,R by physical side). If the Flip's uids differ, stereo will be mirrored or mono and I need your mapping.
4. `journalctl -u hp-elitebook-x-g2i-camera -b | grep -o 'engine=[a-z-]*' | tail -1` → want `engine=camhal` on the **first** app open after boot, not `softisp-fallback`
5. Then the usual: `pactl list cards | grep 'Active Profile'` (HiFi), Chrome/Firefox/OBS, speaker L/R by ear.

If (3) differs, send the `ls` output and which physical speaker each uid drives — mute three at a time via `amixer -c0 cset name='tas2783-N Speaker Volume' 0`.

Known gaps: PR is CONFLICTING against master (my side, working on it); `hp-elitebook-x-g2i-camhal-runtime` depends on `jsoncpp` unversioned, so rebuild it after any jsoncpp major bump.

cyberKrakn:
@meirdick Pushed and booted `hp-elitebook-x-g2i` @ `038c640` on `linux-ptl 7.2.3.arch1-4-ptl`, reference `D4VJ0UA#ABA` (board `8E86`, same as yours).

**Camera: working.** `engine=camhal` on first open, no softisp fallback. Confirmed live via `ffplay -f v4l2 /dev/video50` — clean image, stable service (no restart loop).

**Audio: topology and mapping correct, calibration doesn't apply.**

- `journalctl -k -b`: `sof-sdca-2amp-id2.tplg` loaded (not 1amp) ✓
- `ls /sys/bus/soundwire/devices/`: `sdw:0:2:0102:0000:01:{9,a,c,d}` — matches your reference order ✓
- `calib data update done` count: **0**, want 4 ✗ — but the EFI var is there: `CALI_DATA-02f9af02-7734-4233-b43d-93fe5aa35db3` exists, 116 bytes, same GUID your patch checks.
- Only amp uid `0x9` (right woofer) fails:
  ```
  slave-tas2783 sdw:0:2:0102:0000:01:9: Initialization not complete
  slave-tas2783 sdw:0:2:0102:0000:01:9: ASoC error (-22): at snd_soc_pcm_component_pm_runtime_get()
  ```
  repeating on a loop, no matching errors on a/c/d.
- Net effect is worse than "one amp silent" — `pactl list cards` shows only `off` and `pro-audio` profiles, no `HiFi` at all, card unusable. Confirmed with a raw ALSA open, bypassing PipeWire/UCM entirely:
  ```
  $ speaker-test -D hw:0,2 -c2 -t wav -l1
  Playback open error: -22,Invalid argument
  ```
  Same -22 as the kernel log, so it's not a UCM/PipeWire routing issue — the PCM itself won't open.

Looks like 0x9 failing blocks pm_runtime resume for the whole component chain. Let me know what else you want pulled — happy to add dyndbg on the calibration read path if that helps narrow why the var isn't being consumed despite matching the GUID.

meirdick:
@cyberKrakn fixed on `linux-ptl 7.2.3.arch1-5`, branch pushed (`d881977`).

Your `0x9` hadn't re-attached to the bus when the first PCM opened. Runtime resume waited 5 s, failed, and took the whole card with it. Patch `0067`: no wait, stream starts without a missing amp, it joins on the next stream. Calibration messages are now visible in a plain journal.

Retest and send:
```
journalctl -k -b | grep -E 'factory calibration|not attached'
cat /sys/bus/soundwire/devices/sdw:0:2:0102:0000:01:*/status
```
Expect HiFi and sound. Right woofer may sit out the first stream.

cyberKrakn:
@meirdick Confirmed on `linux-ptl 7.2.3.arch1-5-ptl`, board `8E86` (`D4VJ0UA#ABA`), clean boot.

`journalctl -k -b | grep -E 'factory calibration|not attached'`:

```
slave-tas2783 sdw:0:2:0102:0000:01:9: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:a: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:c: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:d: factory calibration applied (4 speakers in UEFI)
```

`not attached`: 0 lines.

`cat /sys/bus/soundwire/devices/sdw:0:2:0102:0000:01:*/status` -> `Attached` on all four (`9`, `a`, `c`, `d`).

- `sof-sdca-2amp-id2.tplg` loaded (not `1amp`); `factory calibration applied` count is 4.
- `0x9` is up and calibrated — no `-22`, no `snd_soc_pcm_component_pm_runtime_get()` loop.
- `pactl list cards` -> `Active Profile: HiFi`; speaker sink `RUNNING`; all five sinks present.
- Camera still good on the same boot: `engine=camhal` on the first open, no fallback.

Dropping the wait and letting the stream start without the amp is exactly right — the card comes up regardless of bus timing, and the amp joins when it's ready. The `dev_info`/`dev_warn` calibration messages are the other half of the win: this was diagnosable from a plain `journalctl -k -b` with no dyndbg, which is what let me confirm it in one boot instead of a probe session.

Thanks for the fast turnaround and for sticking with it across the two boards — genuinely appreciate the work you put into this.

One note for the record: `hp-elitebook-x-g2i-camhal-runtime` is still `pkgrel 1` here (camera works fine against the current `jsoncpp`), so I'll pick up `pkgrel 2` on the next package sync rather than hand-building it.

meirdick:
Superseded by #480 (same work, rebased onto linux-omarchy 7.2.5). 
@cyberKrakn Omarchy 4.0.4 moved the default kernel to `linux-omarchy` and
dropped `linux-ptl`, so I have rebased the whole stack onto it. New branch
`hp-elitebook-x-g2i-omarchy`: same fixes, 11 patches instead of 20 (nine are
already in Krzysztof's base), everything applies clean on 7.2.5-4, and the
audio and camera packages now depend on `linux-omarchy`.

To test, build `pkgbuilds/linux-omarchy` from that branch (it is a full kernel
build, ~2.5 h on 8 threads; `linux-omarchy 7.2.5-4.1` sorts above the
published `-4`), install it with its headers, and add
`linux-omarchy linux-omarchy-headers` to `IgnorePkg` so the next `omarchy
update` does not replace it. DKMS rebuilds the camera modules on install.
Then the same checks as before:

```
journalctl -k -b | grep -E 'factory calibration|not attached|splitting into 2 PDIs'
cat /sys/bus/soundwire/devices/sdw:0:2:0102:0000:01:*/status
pactl list cards | grep 'Active Profile'
```

Expect HiFi and stereo; the right woofer may sit out the first stream if it
re-attaches late. One thing new to this kernel: it runs the IOMMU in DMA
remapping mode by default. If the camera does not come up, please send
`cat /sys/kernel/iommu_groups/*/type | sort | uniq -c` and try `iommu=pt` on
the command line.

This PR will be closed in favour of the new one once you confirm.

ryanrhughes:
Superseded by #480 from the same author, which carries the newer version of this work.

PR #131, opened 2026-08-11
Update linux-ptl to 7.1.8

spencerbull:
## Summary

- update `linux-ptl` to the Arch Linux 7.1.8-arch1 source and configuration
- retain only the hardware-scoped Panel Replay/ALPM and DisplayID Adaptive-Sync patches
- drop the obsolete SDCA, early-transport, and TRANS_PUSH patches

## Why

Stock Arch still disables Panel Replay on the Dell XPS 14 DA14260 with the LGD OLED panel. The retained workaround replaces only that exact disable quirk, exits Panel Replay for all relevant frontbuffer activity, and re-enters it after 50 ms of idle time. Handling both atomic primary-plane flips and cursor updates is required to keep cursor and video motion smooth.

The DisplayID parser keeps the panel's 20-120 Hz Adaptive-Sync range visible to DRM. The upstream files touched by the three retained patches are unchanged between 7.1.7 and 7.1.8, so the validated patches carry forward without a semantic rebase.

## Validation

- source SHA-256/BLAKE2 checks and upstream PGP signatures passed
- all three patches applied cleanly to 7.1.8-arch1
- `./bin/build --package linux-ptl`: 1 built, 0 failed
- booted `7.1.8-arch1-1-ptl` on the target DA14260/LGD panel
- exact hardware workaround applied; Panel Replay Selective Update reached source SLEEP and sink active/RFB state
- 120 fps mpv test ran at 120.000 Hz with zero dropped, delayed, or mistimed frames
- trace confirmed every relevant frontbuffer flush exited Panel Replay and re-entry occurred after approximately 50.7-50.8 ms
- eDP DPMS off/on passed
- no PSR timeout, DSB poll, FIFO underrun, GPU reset/hang, or AUX errors were observed


meirdick:
@spencerbull incoming for HP Elitebook g2i - https://github.com/omacom-io/omarchy-pkgs/pull/124, doing a rebuild on 7.1.8 now.

PR #137, opened 2026-08-12
Add fresh-editor-bin 0.4.9 from AUR

sinelaw:
Adds `fresh-editor-bin`, a terminal text editor and full-featured IDE.

Site: [getfresh.dev](https://getfresh.dev) · Source: [sinelaw/fresh](https://github.com/sinelaw/fresh) · [v0.4.9](https://github.com/sinelaw/fresh/releases/tag/v0.4.9) · AUR: [fresh-editor-bin](https://aur.archlinux.org/packages/fresh-editor-bin)

### What it is

A terminal editor that does not stop at editing:

- LSP with code actions, goto-definition, references, hover and format-on-save
- Fuzzy file finder, project-wide ripgrep search with live results, file explorer with git status
- Magit-style git review: split-panel diffs, hunk-level stage/unstage/discard, log viewer, git grep
- Multiple cursors and block selection, vim mode, keyboard macros, regex find and replace
- Integrated terminal, SSH remote editing with auto-reconnect, and an orchestrator dock that treats git worktrees as separate workspaces
- Coding agents (claude, codex, opencode, aider) per workspace, persisting across restarts
- Stays responsive on multi-GB files; works as `$EDITOR` via `--wait`

### Packaging

AUR-backed, so `bin/sync-aur` tracks upstream releases with no manual work. Disclosure: I maintain both the editor and the AUR package, so version bumps land there promptly.

The package pulls the prebuilt release tarballs and supports both architectures.

### Verification

Built locally with `bin/build` on both arches at 0.4.9-1:

- `fresh-editor-bin-0.4.9-1-x86_64.pkg.tar.zst`
- `fresh-editor-bin-0.4.9-1-aarch64.pkg.tar.xz` (QEMU)

Source checksums validated against the GitHub release tarballs, no packaging issues reported on either arch.

### Open questions

- **Release ring.** Left unset, so it builds for edge and reaches stable through `bin/repo migrate`. Happy to add `"release_ring": "fast"` if you'd rather it go straight to stable.
- **Install receipt.** The PKGBUILD writes an install receipt marked `channel = "aur-bin"`, which is inaccurate for users installing from the Omarchy repo with pacman. It only affects how the editor talks about updates. I'll fix it upstream rather than patch it here, so it survives future AUR syncs.

### DIFF
--- a/pkgbuilds/fresh-editor-bin/.omarchy/package.json
+++ b/pkgbuilds/fresh-editor-bin/.omarchy/package.json
@@ -1,0 +1,4 @@
+{
+  "source": "aur",
+  "upstream_commit": "6b5782203ab535c451ddce02febae47db04b3a5b"
+}
--- a/pkgbuilds/fresh-editor-bin/LICENSE
+++ b/pkgbuilds/fresh-editor-bin/LICENSE
@@ -1,0 +1,117 @@
+GNU GENERAL PUBLIC LICENSE
+Version 2, June 1991
+
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.
+51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
+
+Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
+
+Preamble
+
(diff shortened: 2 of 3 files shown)


dhh:
Appreciate the suggestion, but not something I could see using or has a big following. We can't host everything from AUR on here.

PR #147, opened 2026-08-15
Install omarchy-network-portal-watch.service

scottjones:
basecamp/omarchy#6807 adds `omarchy-network-portal-watch.service`, a session watcher that announces networks behind a captive portal. `omarchy-settings` installs user units explicitly and `enable-user-units.sh` enables the watcher on first run, so the packages need this line before that PR lands — without it first run aborts under `set -euo pipefail`.

One identical line added to both `omarchy-settings` and `omarchy-settings-dev`, right beside the other user units.


scottjones:
Fresh coordination note: this packaging PR and https://github.com/omacom/omarchy/pull/6807 are the two halves of the captive-portal watcher feature. #6807 adds the watcher command, user service, and activation; this PR installs that service into systemd’s user-unit directory for both omarchy-settings variants. Both are still needed to ship the feature.

The service is currently absent from both quattro and the stable source pinned by omarchy-settings, so the unconditional install here cannot safely land ahead of the source change. Proposed sequencing: make the install conditional on the service file being present, land that packaging support first, then land #6807 and build/publish packages from source containing the service. The conditional-install adjustment has not been pushed yet.

Separately, the September 21 CI run failed before package building: `git status --short | head` under `pipefail` exited 141 in both jobs. That workflow failure is separate from the source/packaging dependency above.

PR #150, opened 2026-08-15
Add Dell XPS 13 sidecar amplifier workaround

spencerbull:
## Summary

- add a removable `dell-xps13-sidecar-amps` x86_64 package for the Dell XPS 13 DX13260 with subsystem ID `1028:0e53`
- select the existing `SOC_SDW_SIDECAR_AMPS` driver path through `snd_soc_sof_sdw quirk=65536`
- rebuild Limine UKIs or mkinitcpio presets on install, upgrade, and removal
- migrate the exact legacy local override while preserving any modified user-owned version
- provide a root-only, idempotent `dell-xps13-sidecar-amps-apply` command so system integrations can directly observe cleanup and rebuild failures
- document the Linux 7.2 removal gate and install package documentation and licensing

## Why

Linux commit [`efd80de2de9d`](https://github.com/torvalds/linux/commit/efd80de2de9d06ddf0eee55ca11b04e39bfc7cd8) adds this exact machine to the SOF SoundWire quirk table with `SOC_SDW_SIDECAR_AMPS`. It first appears in Linux 7.2-rc5. Arch currently ships Linux 7.1.8, so the XPS 13 sidecar amplifiers remain disabled without a temporary override.

This package provides an independently removable compatibility layer. Once every bootable kernel contains the upstream commit, removing the package deletes the modprobe drop-in and rebuilds the boot images. It does not replace the kernel, firmware, or speaker tuning.

Version `1.0.0-2` exposes the apply step as an idempotent command. Pacman reports package scriptlet failures but libalpm does not propagate them through the transaction status, so Omarchy can invoke this command after installation and fail its migration reliably if legacy cleanup or boot-image rebuilding fails.

### Packaging

- rebased onto current `omacom-io/omarchy-pkgs:master`
- direct `makepkg --cleanbuild --clean --force`
- repository-native `bin/repo build --package dell-xps13-sidecar-amps`: 1 built, 0 failed
- x86_64 selected and aarch64 excluded by repository dry runs
- package payload, source checksums, Bash syntax, ShellCheck 0.11, and `git diff --check`
- eight lifecycle fixture cases covering the root guard, exact legacy migration, preservation of modified files, hash and removal failures, masked Limine per-kernel errors, mkinitcpio fallback, and rebuild failure propagation
- cross-repository Omarchy retry/failure matrix confirms explicit apply failures stay observable after Pacman registers the package
- independent findings-first review completed clean after resolving the ALPM scriptlet-status boundary

### Live Dell XPS 13 DX13260

Hardware enablement was previously tested on SKU `0E53` running Arch Linux `7.1.8-arch1-3` with package `1.0.0-1`:

- installed with zero altered package files
- Limine UKI rebuilt successfully and the machine rebooted normally
- loaded module parameter changed from `-1` to `65536`
- both left and right CS35L56 amplifiers bound successfully
- exact `1028:0e53` DSP firmware loaded and calibration applied to both amplifiers
- low-level sweep targeted the physical internal SOF/SoundWire speaker sink; both amplifiers were active and no playback-time audio errors or failed units appeared

The `1.0.0-2` change is limited to making the existing cleanup/rebuild operation directly callable and observable; it was not reinstalled on the live device during this review. A live uninstall-and-reboot remains intentionally skipped because it would disable the now-working amplifiers; removal and rebuild behavior is covered by the lifecycle fixture.

### DIFF
--- a/pkgbuilds/dell-xps13-sidecar-amps/.omarchy/package.json
+++ b/pkgbuilds/dell-xps13-sidecar-amps/.omarchy/package.json
@@ -1,0 +1,3 @@
+{
+  "source": "local"
+}
--- a/pkgbuilds/dell-xps13-sidecar-amps/LICENSE
+++ b/pkgbuilds/dell-xps13-sidecar-amps/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Spencer Bull
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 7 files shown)


sharms:
I tried this last night and WOW huge difference - the XPS 13 speakers are really good

areen-dev:
Thanks it worked and the difference is of day and night🙌

PR #163, opened 2026-08-17
Fix SoundWire audio on ASUS ExpertBook B9406

argrig666:
## What

Add a hardware-scoped `linux-ptl` patch for the ASUS ExpertBook B9406CAA and bump the package release.

The laptop firmware advertises an unattached RT722 at SoundWire address `0x000330025d072201` beside the real CS42L43. Linux 7.1's function-topology builder creates two `SDW3-Playback-SimpleJack` links, hits `-EEXIST`, and aborts `sof_sdw` registration, leaving no ALSA card.

The patch remaps only that phantom address to zero for ASUS board name `B9406CAA`, so SoundWire ignores it and the real CS42L43 plus both CS35L56 amplifiers enumerate normally.

## Live hardware validation

Tested on ASUS ExpertBook B9406CAA, PCI audio subsystem `1043:15e4`, kernel `7.1.8-arch1-3`:

- Before the quirk: duplicate `SDW3-Playback-SimpleJack`, `sof_sdw` probe failed with `-12`, no ALSA card.
- With the equivalent DKMS quirk: `sof-soundwire` registers, both CS35L56 DSPs load OEM firmware and calibration, and PipeWire exposes the HiFi Speaker sink.
- The current Arch `alsa-ucm-conf 1.2.16.1` and `linux-firmware-cirrus` already contain the needed UCM and firmware, so no userspace files are bundled here.

## Validation

- `makepkg --nobuild --nodeps --skippgpcheck` completes and applies all four kernel patches.
- SHA-256 and BLAKE2 checksums pass through `makepkg --verifysource`.
- `drivers/soundwire`, including `dmi-quirks.o` and `soundwire-intel.ko`, compiles and links against the prepared package source.
- `bin/omarchy-pkgs self-test` passes.

A coordinated `basecamp/omarchy` PR will select this kernel only on the already-supported PTL XPS models and the B9406.

### DIFF
--- a/pkgbuilds/linux-ptl/0030-soundwire-ignore-ghost-rt722-on-asus-b9406.patch
+++ b/pkgbuilds/linux-ptl/0030-soundwire-ignore-ghost-rt722-on-asus-b9406.patch
@@ -1,0 +1,52 @@
+From: argrig666 <66850845+argrig666@users.noreply.github.com>
+Subject: [PATCH] soundwire: ignore ghost rt722 on ASUS ExpertBook B9406CAA
+
+The ASUS ExpertBook B9406CAA firmware advertises a phantom Realtek
+rt722 on SoundWire link 3 alongside the real CS42L43 codec. Linux 7.1's
+function-topology builder creates the same SDW3-Playback-SimpleJack link
+for both devices, then fails the sof_sdw probe with -EEXIST. No ALSA card
+is registered as a result.
+
+Remap the phantom rt722 address to zero so the SoundWire core ignores it.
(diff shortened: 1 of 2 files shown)


rubas:
The fix is now in mainline as 90af3209742d ("soundwire: dmi-quirks: Disable ghost Realtek on Asus Expertbook"), queued for v7.3: https://github.com/torvalds/linux/commit/90af3209742db61a7f9d7d054a16165818cfc6d8

Since master is on 7.2.3 now, this could become a straight backport of that commit. Happy to test a rebuilt package on a B9406CAA.

PR #171, opened 2026-08-19
Build aarch64 packages on native ARM64 runners

scottjones:
Building aarch64 packages on an x86_64 host requires QEMU. This adds a manually dispatched **Build aarch64 Packages** workflow on `ubuntu-24.04-arm`, with a native architecture check and unsigned package artifacts retained for 14 days.

Choose `edge`, `rc`, or `stable`, and optionally provide space-separated recipe names. Package channel and release-ring policy still applies; ordinary packages build on `edge`. Targeted builds must include any required recipes from this repository; package dependencies are not automatically added to the selected set. A full build has a six-hour timeout. A run that produces no package files fails visibly, and the artifact contains only package archives.

To publish, download and extract the artifact **on the repository host** into `build-output/<mirror>/aarch64`, then run:

```bash
bin/upload-prebuilt --arch aarch64 --mirror <mirror>
```

Use the same mirror as the build. This signs, promotes, updates the repository database, and syncs under one release lock.

The workflow complements the scheduled architecture support from #277. It does not enable scheduled ARM publishing; the checked-in `PUBLISHED_ARCHES` default remains `x86_64`. The PR changes only the workflow and README. #223 adds reusable workflow calls and a repository artifact to the same workflow file and will need to reconcile its changes after this lands.

Validation:
- All four repository self-test commands passed on the refreshed branch.
- Workflow YAML parsed successfully; embedded shell scripts passed `bash -n`; `git diff --check` passed. The summary script was exercised with empty output (failure) and a package file (success).
- Native ARM smoke builds at `3cc6f83` passed: [`tzupdate` on edge](https://github.com/scottjones/omarchy-pkgs/actions/runs/34148331444) and [`symfony-cli` on rc](https://github.com/scottjones/omarchy-pkgs/actions/runs/34148332994), each with one package built and zero failures. Downloaded both artifacts and verified `.PKGINFO` reports `aarch64` and their executables have ELF machine type AArch64.
- [PR self-tests](https://github.com/omacom/omarchy-pkgs/actions/runs/34148336402) passed at the same commit.

A full repository build and production publication were not exercised.


scottjones:
There are now eight open aarch64 PRs on this repo and we're starting to collide. Posting a map here since this PR is the one several of the others already reference.

### How they fit

| Layer | PR | Author |
|---|---|---|
| Native ARM64 runner | **#171** (this one) | @scottjones |
| Reusable repo artifact / `omarchy.db` publication | #223 | @birkskyum |
| Recipes + builder toolchain | #240 | @riverscn |
| `arch=()` flips | #195, #197, #244 | @oceanapplications, @andrewpatrickflynn |
| Platform bring-up | #221 (Snapdragon fw), #222 (Limine) | @birkskyum |

These are complementary rather than competing. #223 says so directly — it carries this branch's commits so it can be exercised today, and plans to rebase to only its follow-up commits once this merges. #240 is the mirror image: its `docs/aarch64-follow-up.md` explicitly defers "native AArch64 production scheduling", states it "does not add an emulated GitHub Actions job", and asks that release acceptance be performed on a native AArch64 builder. That is exactly what this PR adds, and it isn't emulated — `ubuntu-24.04-arm`, free for public repos.

### Where we actually conflict

Until a few minutes ago three PRs flipped `arch=()` on the same two packages, under three different patch filenames for an identical change:

| Package | ~~#171~~ | #195 | #240 |
|---|---|---|---|
| `tensaku` | ~~`build-for-aarch64.patch`~~ | `add-aarch64-arch.patch` | `aarch64.patch` |
| `tzupdate` | ~~`build-for-aarch64.patch`~~ | `add-aarch64-arch.patch` | `aarch64.patch` |

**I've removed those from this PR** (see below), so what remains is #195 and #240, which also overlap on `asdcontrol`, `github-copilot-cli`, `hyprland-preview-share-picker`, `omasnap`, `qmk-hid`, `symfony-cli` and `v4l2-relayd`.

The diffs agree — `arch=('x86_64')` → `arch=('x86_64' 'aarch64')` in every case. It's a naming and ownership problem, not a design disagreement.

### Proposed sequencing

1. **Done —** I've dropped the tensaku/tzupdate changes from this PR. It's now two commits and two files (`.github/workflows/build-aarch64.yml` and a README note), +154/-1, with no file in common with #195 or #240. The remaining overlap with #223 is the workflow itself, which is the dependency rather than a conflict.
2. Merge this, so #223 can rebase down to its own commits.
3. Let #240 own the `arch=()` flips and settle on its `aarch64.patch` convention, with #195/#197/#244 converging on it.

Happy to reorder if maintainers would rather take #240 first — the point is mainly that these want an order rather than parallel resolution of the same files.

birkskyum:
@scottjones if you can land this one, i can rebase the prs i have that build on top.

scottjones:
Thanks — items 1, 2 and 3 are fixed in 7d50d88. All three were comment-only, so the workflow's behaviour is unchanged; what changed is that the comments no longer describe things that do not work.

**1. Publish recipe (medium).** Confirmed, and the diagnosis was exactly right. `bin/promote-build:86,123,173` skips `omarchy-build.db*` on every path that moves a file into the published tree, `bin/repo update` is the only caller of `build/update-repo.sh`, and its `repo-add` at line 66 is what writes `omarchy.db`. `bin/sync-repo:235-237` then uploads `--include "omarchy.*"` out of `$REPO_DIR` — so with no `update` step there is simply no database file there to upload, and the packages land unresolvable. The recipe now reads:

```
bin/repo sign    --arch aarch64 --mirror <mirror>
bin/repo promote --arch aarch64 --mirror <mirror>
bin/repo update  --arch aarch64 --mirror <mirror>
bin/repo sync    --arch aarch64 --mirror <mirror>
```

with `--mirror` threaded through all four for the `MIRROR=edge` default you flagged, a note on where to unpack the artifact, and a short paragraph explaining why `update` is not optional so the next person does not trim it back out.

**2. Batching advice (low).** Confirmed. `build/build.sh:511` only counts a dependency when it is also in `PACKAGES_TO_BUILD`, and `build/build.sh:60-61` gates the `[omarchy]` repo on a database that does not exist on a clean runner, so a batch has to be dependency-closed. `pkgbuilds/omarchy/PKGBUILD:32` pins `omarchy-settings=${pkgver}` and is the obvious way to trip it. The comment no longer offers the input as a way to split a full build; it says what the input is for and what a batch has to satisfy.

**3. Log upload (low).** Confirmed, and slightly worse than described — not only does this workflow call `bin/build` directly while `logs/` is written only by `bin/repo:26,39`, but `build/build.sh` writes no log file anywhere; it is all stdout. There was nothing for the step to collect under any arrangement, so I removed it rather than repointing it. The job log is the diagnostic.

**4. `chmod -R 777` (low).** Left as is, deliberately. Your own analysis is the reason: `ubuntu-24.04-arm` gives each job a fresh single-tenant VM that is destroyed afterwards, so this is not a privilege boundary, and the mode has to be opened before `make_dir_writable()`'s recursive `chown` for the uid 1000/1001 split to work at all. `setfacl` would be tidier in the abstract, but #240 rewrites `make_dir_writable()` to `chmod -R a+rwX` and #223 does `chmod -R a+rwX` in its own copy of this workflow — so switching to ACLs here would make this the only one of the three doing something different, for no gain on an ephemeral runner. Happy to revisit if `make_dir_writable()` gets fixed properly, which would let this step disappear entirely.

On the sequencing note at the end: agreed, and thanks for catching it — I had #223 as stacked on this branch, and it is a competing rewrite of the same file rather than a follow-up. That is worth settling before either merges. cc @birkskyum, since it changes what "rebase on top" means for #223 specifically — #221 and #222 are unaffected.

birkskyum:
@ryanrhughes , this contriubtion has become quite a bottleneck - it's holding back an entire cascade of other PRs related to improving ARM support.

scottjones:
@birkskyum This is refreshed against current master and ready for another look. The README conflicts are resolved, `rc` is available, and the publishing instructions now point to `bin/upload-prebuilt` on the repository host. I also removed the permission workaround that master now handles and made empty builds fail visibly.

At `3cc6f83`, the PR self-tests pass, and both native ARM smoke builds passed: [tzupdate on edge](https://github.com/scottjones/omarchy-pkgs/actions/runs/34148331444) and [symfony-cli on rc](https://github.com/scottjones/omarchy-pkgs/actions/runs/34148332994). I downloaded both artifacts and verified the package metadata and executable architecture are AArch64. GitHub reports no merge conflicts.

This still only changes the workflow and README. My preference is to land this first, then rebase #223 so its reusable calls and repository artifact build on the merged workflow. It is ready for maintainer review and merge.

birkskyum:
Thanks, Scott. Landing #171 first works for me. I've updated #223 to be ready to merge in after this.

birkskyum:
@ryanrhughes, could you take a look at this PR when you have a chance? It complements the ARM pipeline work you reviewed in #277 by adding manually dispatched native ARM builds. Scott and I have coordinated how [#223](https://github.com/omacom/omarchy-pkgs/pull/223) can follow it with reusable workflow calls and a package repository artifact, benefiting both Apple Silicon and Snapdragon support.

On the Snapdragon side, we have Omarchy running on Lenovo Yoga Slim 7x and ASUS Vivobook S15 machines with model-specific setups, plus successful live-USB tests. [omacom/omarchy-iso#129](https://github.com/omacom/omarchy-iso/pull/129) uses that repository to build the installer.

The refreshed workflow has passed native ARM smoke builds and repository self-tests. Once this lands, I can rebase #223 onto it and add the reusable workflow calls and ARM64 package repository artifact.

birkskyum:
I’ve [updated the Dragon ISO](https://github.com/omacom/omarchy-iso/commit/af8b43615044fe7d77e2cb2dfbb5ca643ef17723) to consume the published ARM repositories directly, so this is no longer needed from for snapdragon support.

PR #176, opened 2026-08-20
Add hey-cli, built from source

omarchybot:
Omarchy had no package for the HEY CLI, so getting `hey` onto a machine meant cloning basecamp/hey-cli and running `make install` — including on the coordinator, where omabot's triage report is mailed out through it.

There is nothing released to package. basecamp/hey-cli has no tags and no releases, and the AUR's `hey-cli` is a name reservation: its source URL points at a `v0.0.1` tarball that GitHub answers with a 404, so an `.omarchy` AUR sync would import a PKGBUILD that cannot build. This builds from a pinned commit instead, the way herdr and omatrack do, and keeps the name, provides/conflicts and install paths the AUR entry reserved because those are upstream's own intent.

`pkgver` is `0.0.0.r132`, with `0.0.0` standing in for the release that has not happened. It sorts below every version upstream might plausibly tag — the reserved `0.0.1` included — so the first real release is an upgrade rather than a downgrade. When that release lands, point `_commit` at the tag and drop the `.rN`; or hand the package back to the AUR once upstream's goreleaser starts publishing there, which is where basecamp-cli already gets its PKGBUILD.

Edge only: a package with no upstream release has not earned the fast ring.

### DIFF
--- a/pkgbuilds/hey-cli/.omarchy/package.json
+++ b/pkgbuilds/hey-cli/.omarchy/package.json
@@ -1,0 +1,3 @@
+{
+  "source": "local"
+}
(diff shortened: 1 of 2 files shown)


dhh:
Close in favor of managing this through mise.

PR #178, opened 2026-08-20
Package yt-dlp ahead of Arch's extra

omarchybot:
yt-dlp breaks the moment a site changes its player, and upstream ships a fix within days. Arch's `extra` is on **2026.07.04** as of this PR while upstream is on **2026.08.19** — six weeks of extractor fixes that Omarchy users do not have.

The PKGBUILD is Arch's, with only `pkgver` and `sha256sums` changed. Keeping it a two-line delta means it stays cheap to diff against theirs, and cheap to drop entirely if Arch ever keeps pace. Upstream still pins `yt-dlp-ejs==0.8.0`, so `depends` needed no change.

`.omarchy/upstream.sh` keeps it current without anyone watching:

- the version comes from the `/releases/latest` redirect rather than the GitHub API, which is rate limited per IP and would fail the sync on a busy build host
- the checksum is read out of the `SHA2-256SUMS` that yt-dlp signs and publishes beside every release, so an update is two small requests instead of a 5.7 MB download

It builds in the **fast ring** deliberately. A package that shadows `extra` has to reach stable on its own; waiting for a promote would leave stable users on an older yt-dlp than `extra` would have handed them.

One thing to know before merging: this only becomes reachable once `[omarchy]` is ordered above `[extra]`. pacman stops at the first repository carrying a name and never compares versions across the rest, so under today's ordering this package is built and published but never installed. The ordering change is a separate PR against `omarchy`, and it should not land before the deadweight cleanup in #177.

### DIFF
--- a/pkgbuilds/yt-dlp/.omarchy/package.json
+++ b/pkgbuilds/yt-dlp/.omarchy/package.json
@@ -1,0 +1,4 @@
+{
+  "source": "local",
+  "release_ring": "fast"
+}
--- a/pkgbuilds/yt-dlp/.omarchy/upstream.sh
+++ b/pkgbuilds/yt-dlp/.omarchy/upstream.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# yt-dlp ships a signed SHA2-256SUMS beside every release, so an update costs two
+# small requests instead of a 5.7 MB download. The version comes from the
+# /releases/latest redirect rather than the API, which is rate limited per IP and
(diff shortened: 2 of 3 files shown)


ryanrhughes:
Closing: Arch's extra caught up and now ships yt-dlp 2026.08.19, the same version this PR carried, so there is nothing for the Omarchy repo to get ahead of. Thanks for the work on the upstream hook; if Arch falls behind again we can revisit.

PR #184, opened 2026-08-21
omarchy-nvim: Add iron.nvim configuration for interactive REPL

apurbapokharel:
### Summary

Adds [`iron.nvim`](https://github.com/Vigemus/iron.nvim) plugin configuration to the `omarchy-nvim` package to provide an out-of-the-box interactive REPL (supporting Python/IPython and Shell with `# %%` code cell blocks).

### Changes
* Added `pkgbuilds/omarchy-nvim/lua/plugins/iron.lua` with:
  * Dynamic fallbacks (`ipython` -> `python3`, `zsh` -> `$SHELL`/`bash`).
  * On-demand lazy loading via `cmd` and `keys`.
  * Which-key group descriptions under `<leader>r`.
  * Dedicated terminal mode escape mapping (`<Esc>` -> normal mode).

### Companion PR
* `basecamp/omarchy`: https://github.com/basecamp/omarchy/pull/7690

### DIFF
--- a/pkgbuilds/omarchy-nvim/PKGBUILD
+++ b/pkgbuilds/omarchy-nvim/PKGBUILD
@@ -1,6 +1,6 @@
 # Maintainer: Ryan Hughes <ryan@omarchy.org>
 pkgname=omarchy-nvim
-pkgver=2026.8.13
+pkgver=2026.8.23
 pkgrel=1
 pkgdesc="Pre-built LazyVim configuration with cached plugins"
 arch=('any')
--- a/pkgbuilds/omarchy-nvim/lua/plugins/iron.lua
+++ b/pkgbuilds/omarchy-nvim/lua/plugins/iron.lua
@@ -1,0 +1,68 @@
+return {
+  "Vigemus/iron.nvim",
+  cmd = { "IronRepl", "IronRestart", "IronFocus", "IronHide" },
+  keys = {
+    { "<leader>r", desc = "+repl" },
+    { "<leader>rr", "<cmd>IronRepl<cr>", desc = "Toggle REPL" },
+    { "<leader>rR", "<cmd>IronRestart<cr>", desc = "Restart REPL" },
+    { "<leader>rf", "<cmd>IronFocus<cr>", desc = "Focus REPL" },
+    { "<leader>rh", "<cmd>IronHide<cr>", desc = "Hide REPL" },
+    { "<leader>rs", desc = "+send" },
+    { "<leader>rsl", "<cmd>lua require('iron.core').send_line()<cr>", desc = "Send Line" },
+    { "<leader>rsf", "<cmd>lua require('iron.core').send_file()<cr>", desc = "Send File" },
+    { "<leader>rsp", "<cmd>lua require('iron.core').send_paragraph()<cr>", desc = "Send Paragraph" },
+    { "<leader>rsu", "<cmd>lua require('iron.core').send_until_cursor()<cr>", desc = "Send Until Cursor" },
+    { "<leader>rsb", "<cmd>lua require('iron.core').send_code_block(false)<cr>", desc = "Send Code Block" },
+    { "<leader>rsn", "<cmd>lua require('iron.core').send_code_block(true)<cr>", desc = "Send Code Block & Move Next" },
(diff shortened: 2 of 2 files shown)


apurbapokharel:
A short video showing the proposed REPL-based system. Would allow Python notebook capabilities from nvim.

https://github.com/user-attachments/assets/9cb29f43-8c67-4e71-b03b-c724a25fd904

apurbapokharel:
Thanks for the thorough review — all five items are now addressed.

**On this PR (e0c003a):**

- **`pkgver` bump** — bumped to `2026.8.23` so `check_needs_build` picks up the change and `bin/repo release` actually rebuilds omarchy-nvim.
- **`PYTHON_BASIC_REPL`** — added `env = { PYTHON_BASIC_REPL = "1" }` to the python repl definition, matching the annotated line in iron.nvim's own README, so the Python 3.13+ PyREPL no longer compounds the indentation of sent blocks.
- **Comment filter** — removed the `^%s*#` filter entirely; it was silently dropping lines inside triple-quoted strings and docstrings. The definition now uses iron's stock `bracketed_paste_python` (kept explicitly, since a custom `repl_definition` entry without `format` falls back to `repldef.open`/`close`, which aren't set here).
- **`<esc>` scope** — the terminal-mode mapping is no longer global. `repl_filetype` now returns `"iron"` and the mapping is created buffer-local via a `FileType iron` autocmd, so a plain `:terminal` keeps its normal Escape behavior.
- **Keymap cleanups** — dropped the duplicate visual `<leader>rsv` (kept `<leader>rs`), and added the `+clear` group label for `<leader>rc`.

**On the companion PR (basecamp/omarchy#7690, 2ff81e74):**

- Added `migrations/1787498986.sh`, following the `remote_clipboard.lua` migration pattern: it installs `iron.lua` from `/usr/share/omarchy-nvim/config/lua/plugins/` into an existing user's `~/.config/nvim`, guarded on both the config dir and the packaged source existing. Existing users now get the keybindings the manual documents.
- Updated the hotkeys manual and the keybindings menu to say the Escape mapping applies in the REPL terminal specifically, matching the new scoping.

Understood that whether Omarchy ships a REPL at all is the maintainer's call — these changes just make sure the PR is technically ready either way.

apurbapokharel:
Thanks for the re-review — both new findings are addressed, and I took your `repl_filetype` suggestion as proposed.

**On this PR (91306d6):**

- **The six broken keybindings** — fixed exactly the way you outlined: the buffer-local terminal-mode `<esc>` mapping is now created inside `repl_filetype`, which receives the REPL's `bufnr`, and the function returns the source `ft` again so `ll.get_buffer_ft` resolves inside the REPL window. The `FileType iron` autocmd is gone. This keeps both properties at once: Escape stays scoped to iron's REPL buffers (plain `:terminal` untouched), and `rr`/`rh`/`rf`/`rq`/`rcc`/`rx` work from inside the REPL again.

**On the companion PR (basecamp/omarchy#7690, 27d4b09a):**

- **The release-order hole** — the migration no longer guards on the packaged source existing. If omarchy-nvim 2026.8.23 isn't installed yet, `install` fails under the runner's `set -euo pipefail`, the marker is never touched, and the migration retries on the next run — same shape as `1781587663.sh`. That covers the `omarchy-migrate-notify` window you pointed out, where a login can trigger migrations with no package step in front.
- **The user-config overwrite** — the migration now skips when `~/.config/nvim/lua/plugins/iron.lua` already exists, since until this ships any file at that path is one the user wrote themselves. Skipping (rather than backing up and overwriting) matches how `omarchy-nvim-setup`'s `install_config` treats existing config.

Both paths were tested against throwaway `HOME`s: no config dir → clean no-op; config dir with missing source → non-zero exit, marker untouched; user's own `iron.lua` present → untouched, clean exit; normal case → installed 0644.

**Left as-is, for the maintainer's decision:** the `rsb`/`rsn` nil-index on other filetypes (upstream iron bug in `core.lua`), the `<leader>r` namespace question against LazyVim's `editor.refactoring` extra, and the `pkgver` collision with #125 for whoever merges second.

apurbapokharel:
@ryanrhughes, please let me know if there are things  I can do to support this PR. I am unsure if I should solve/resolve the conflicts now that the maintainers have begun working on this.

PR #186, opened 2026-08-22
omamorph: add local media converter and compressor

JoeJoeflyn:
### Summary
Adds the `omamorph` package recipe.

**omaMorph** is a local media converter and compressor for Omarchy (Video, Audio, Images, Documents).

**Repo**: https://github.com/JoeJoeflyn/omamorph

### DIFF
--- a/pkgbuilds/omamorph/.omarchy/package.json
+++ b/pkgbuilds/omamorph/.omarchy/package.json
@@ -1,0 +1,3 @@
+{
+  "source": "local"
+}
--- a/pkgbuilds/omamorph/PKGBUILD
+++ b/pkgbuilds/omamorph/PKGBUILD
@@ -1,6 +1,6 @@
 # Maintainer: JoeJoeflyn <https://github.com/JoeJoeflyn>
 pkgname=omamorph
-pkgver=1.0.2
+pkgver=1.0.3
 pkgrel=1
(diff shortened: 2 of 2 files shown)


JoeJoeflyn:
The missing-tool crash is fixed in **v1.0.3** (tag `v1.0.3` → commit `cd19e33`, pushed to `main`).

**Root cause confirmed.** Reproduced the v1.0.2 crash with a minimal Qt6 probe: `QProcess::errorOccurred(FailedToStart)` → `delete proc` inside the signal emission → `SIGSEGV` (exit 139). Qt resumes on freed memory after destroying the process mid-signal.

**Fix:** `failJob` at `src/backend.cpp:544` now uses `deleteLater()` instead of `delete m_proc`, matching the batch path (`:613`, `:622`). Verified the probe with `deleteLater()` → exit 0, no crash.

**`stderrTail` leak also fixed:** `src/backend.cpp:463` now uses `QSharedPointer<QByteArray>` (ref-counted) instead of a raw `new QByteArray` that was only deleted in the `finished` handler. No leak on failed-start paths.

**Build verified:** clean `qmake6 && make` from `cd19e33` produces `omaMorph` with exit 0.

The PKGBUILD in this PR still points at v1.0.2 — bumping it to v1.0.3 so it picks up the fix. New checksum will follow.

PR #192, opened 2026-08-23
fix(mise-bin): delay upstream releases for 24 hours

jdx:
## Summary

- Hold new mise releases for 24 hours from their GitHub `published_at` timestamp before OPR syncs them.
- Fail closed when the release timestamp is missing or invalid.
- Keep the policy package-specific for now so other OPR packages are unchanged.
- Provide an explicit, package-specific bypass for manually reviewed emergency updates.

## Why

This quarantine window gives critical bugs a day to surface and gives maintainers time to pull a bad release before it reaches Omarchy machines. It also limits immediate distribution if the mise maintainer account or release pipeline is compromised.

The check runs in `mise-bin`'s upstream hook before checksums are fetched or either release ring can build the package. Because upstream sync runs every six hours, a valid release should normally reach OPR 24–30 hours after publication.

## Emergency patches

Scheduled automation does not set the bypass. If a critical mise patch must ship inside the quarantine window, a maintainer can deliberately generate the update locally:

```bash
MISE_BIN_BYPASS_RELEASE_AGE=1 ./bin/sync-upstream mise-bin
```

The resulting PKGBUILD change can then be reviewed and merged through a normal manual PR, preserving an explicit human approval point.

## Testing

- Verified a 23:59:59-old release returns `{}` without fetching its checksums.
- Verified a release exactly 24 hours old is accepted with both architecture checksums.
- Verified the explicit emergency bypass accepts a younger release.
- Verified an invalid `published_at` fails closed.
- Ran `bash -n` and `git diff --check`.

*AI-assisted — Tool: Codex; model: openai/gpt-5; version: unavailable.*

### DIFF
--- a/pkgbuilds/mise-bin/.omarchy/upstream.sh
+++ b/pkgbuilds/mise-bin/.omarchy/upstream.sh
@@ -4,10 +4,29 @@ set -euo pipefail
 repo="jdx/mise"
 release=$(curl -fsSL "https://api.github.com/repos/$repo/releases/latest")
 tag=$(jq -r '.tag_name // empty' <<<"$release")
+published_at=$(jq -r '.published_at // empty' <<<"$release")
 
 if [[ ! "$tag" =~ ^v([A-Za-z0-9._+]+)$ ]]; then
   echo "Latest mise release has an invalid tag: ${tag:-<empty>}" >&2
   exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


jdx:
curious what y'all think of this, I think it'd be good to put a bit of a buffer in place in case I were to get compromised or if there were ever a critical bug in mise so it doesn't speed run out to omarchy users.

I feel like this should be more of a general thing and other packages would benefit from similar logic, I could look into refactoring this to allow for that if there's interest.

PR #195, opened 2026-08-24
Own and build nine ARM-compatible packages

oceanapplications:
## Summary

Enable aarch64 for nine packages that have been built and exercised natively on Arch Linux ARM, while making Omarchy the source of truth for every affected recipe.

The original version used AUR overlay patches. This revision deliberately removes that dependency: the six formerly AUR-synced packages are now `source: local` and update directly from their authoritative vendor feeds. An AUR update can no longer revert aarch64 support or replace their build logic.

| Package | Upstream update source |
|---|---|
| `github-copilot-cli` | declarative npm `latest`, hashing the npm tarball and GitHub changelog |
| `qmk-hid` | declarative Git tag provider |
| `symfony-cli` | declarative GitHub release/checksum-manifest provider |
| `tensaku` | declarative Git tag provider |
| `tzupdate` | declarative Git tag provider |
| `v4l2-relayd` | declarative Git tag provider, retaining and hashing Omarchy's patches |

`asdcontrol`, `hyprland-preview-share-picker`, and `omasnap` were already local recipes and only add `aarch64` to `arch=()`.

## Declarative provider expansion

`sync-upstream` now supports three mutually exclusive JSON provider types:

- `github`: stable GitHub Releases with an upstream checksum manifest (existing behavior)
- `git_tags`: configurable repository, `{pkgver}` tag pattern, and source URL templates
- `npm`: configurable dist-tag and source templates, including `{npm_tarball}`

Remote source templates must expand to HTTPS URLs and redirects remain HTTPS-only. `file:name` entries hash checked-in local sources. Provider configuration is validated and malformed or ambiguous declarations fail closed. These six packages therefore need no bespoke `upstream.sh` files.

## Additional cleanup

- `qmk-hid` advances from 0.1.11 to the current 0.1.13 tag; the source checksum is independently downloaded and pinned.
- `symfony-cli` replaces `git+...#tag` with `sha256sums=('SKIP')` by the release source tarball and Symfony's published SHA-256. `pkgrel` advances to 9 because this changes the source of the existing version.
- `github-copilot-cli` changes its checksum array from BLAKE2 to SHA-256 so `sync-upstream` can update it through the standard hook contract.

## Verification

- All nine PKGBUILDs produce `.SRCINFO` containing both `x86_64` and `aarch64`.
- `makepkg --verifysource` passes for all six migrated recipes.
- Every updater reports no change at the checked-in version.
- The provider suite exercises older fixture versions and verifies selection, source expansion, remote/local hashing, npm metadata, provider exclusivity, and failure behavior.
- `bin/sync-upstream self-test` passes.
- `qmk-hid` 0.1.13 builds and packages successfully on x86_64.
- The original native aarch64 build/install/runtime evidence for all nine packages still applies; the refreshed `qmk-hid` and Symfony source change should receive another native aarch64 build before merge.

Not included: `libfprint-git` and `rustdesk`, whose ARM builds require substantive fixes beyond architecture metadata.


oceanapplications:
For context on where these end up: `pkgs.omarchy.org` currently serves x86_64 only — `stable/aarch64/omarchy.db` and `edge/aarch64/omarchy.db` both 404 — so even once these build for ARM64 there is nowhere to publish them. Opened #199 for that, since it is a publishing step rather than a code change and `bin/build --arch aarch64` already exists.

PR #202, opened 2026-08-24
Quarantine fresh upstream releases via min_release_age in the package manifest

ryanrhughes:
## Summary

Follow-up to #192, twice generalized per review discussion:

1. **The quarantine is a manifest policy.** `.omarchy/package.json` gains optional `"min_release_age": "24h"` (`s`/`m`/`h`/`d` or bare seconds). `bin/sync-upstream` exports it to hooks as `MIN_RELEASE_AGE_SECONDS` and enforces it as a backstop: with a policy set, the reported release must carry `published_at` and be old enough, else it is held (or the sync fails if the age is unprovable). `BYPASS_MIN_RELEASE_AGE=1 ./bin/sync-upstream <pkg>` is the deliberate manual override; automation never sets it.

2. **mise-bin no longer has an upstream.sh at all.** Everything its hook knew was data, now declared in the manifest and handled by a shared GitHub-releases provider (`helpers/upstream-github.sh`):

```json
"upstream": {
  "github": "jdx/mise",
  "checksums": "SHASUMS256.txt",
  "assets": {
    "x86_64": "mise-{tag}-linux-x64.tar.xz",
    "aarch64": "mise-{tag}-linux-arm64.tar.xz"
  }
}
```

The provider walks the release feed (drafts/prereleases excluded), selects the newest release by `vercmp` that has cleared the quarantine — so mise's near-daily cadence can't starve updates (today: ships 2026.8.11, holds 10h-old 2026.8.12) — fails closed on unreadable tags/timestamps, and skips the checksum fetch when nothing is new. Declaring both an `upstream` block and an `upstream.sh` is an error; the hook remains the escape hatch for bespoke feeds (codex's Debian index, tmog's version.txt, t3code's electron-builder manifest — each a future candidate once a fitting convention exists).

Manifest validation covers the new fields; enumeration includes provider-declared packages.

## Testing (live GitHub API, in a worktree)

- `./bin/sync-upstream mise-bin` → 2026.8.11, pkgrel 1 (x64 artifact independently downloaded and hashed earlier: matches)
- `BYPASS_MIN_RELEASE_AGE=1` → 2026.8.12
- pkgver already 2026.8.11 → "No upstream update reported", no checksum fetch
- `min_release_age: 3650d` → all releases quarantined, clean skip
- provider + hook both present → hard error
- stub hook reporting a 14h-old release under a 24h policy → held by the backstop; stub omitting `published_at` → sync fails
- manifest validation rejects an `upstream` block missing `checksums`/`assets`; accepts the real one
- `bash -n` clean on all files; `packages_for_upstream_sync` lists all four packages

cc @jdx


jdx (approved):

jdx:
@ryanrhughes not sure if you saw my comment on #192 but I think it may make sense to generalize this logic if we think we should have this behavior for other packages, would be happy to send a PR if you think it's worth it

ryanrhughes:
@jdx I think that's exactly what we should do. I have an agent reworking this to be more generalized logic and add `min_release_age` to the `.omarchy/package.json` so we could use this logic across more than just mise.

ryanrhughes:
Momus asked for executable proof of the migrated publishing path. `bin/sync-upstream self-test` now covers it offline (fixture-fed fetches, production code paths, including `sync_package` end to end over the checked-in mise-bin metadata with final-PKGBUILD assertions). Output from this branch:

```
================================
Upstream Package Sync
================================
Release selection:
  ok: quarantine falls back past the young v2.0.0
  ok: selected release reports its published_at
  ok: x86_64 checksum via {tag} template and ./ prefix
  ok: aarch64 checksum via {pkgver} template
  ok: no policy selects the newest stable release
  ok: prerelease v1.9.9 and draft v1.9.8 are never selected
  ok: bypass lifts the quarantine
  ok: x86_64 checksum via * binary-mode prefix
  ok: everything quarantined reports no update
  ok: already checked in reports no update
Failure paths:
  ok: invalid published_at fails the sync
  ok: unusable tag fails the sync
  ok: missing aarch64 checksum fails the sync
Quarantine backstop:
  ok: old enough passes
  ok: too young is held
  ok: no policy passes anything
  ok: deliberate bypass passes
  ok: missing published_at is unprovable
Duration parser:
  ok: min_release_age "24h"
  ok: min_release_age "90m"
  ok: min_release_age "2d"
  ok: min_release_age 3600
  ok: min_release_age "600s"
  ok: min_release_age "010h"
  ok: min_release_age "abc"
  ok: min_release_age "24hh"
  ok: min_release_age false
  ok: min_release_age "9999999999"
Manifest validation:
  ok: upstream: false is rejected
  ok: upstream without checksums/assets is rejected
  ok: the real declaration shape is accepted
End-to-end sync_package with the checked-in mise-bin metadata:
  ok: sync_package updates without failures
  ok: the 24h manifest policy holds v2026.9.1 and ships v2026.9.0
  ok: pkgrel resets to 1
  ok: x86_64 checksum lands in the PKGBUILD
  ok: aarch64 checksum lands in the PKGBUILD
  ok: malformed upstream stays discoverable for scheduled runs
  ok: malformed upstream fails the sync instead of skipping

✓ Self-test passed
```

jdx:
awesome, I think that's a great call

PR #213, opened 2026-08-27
Update asusctl to 6.4.0

### LINKED ISSUE #154: rog-control-center crashes on startup — bump asusctl PKGBUILD to 6.4.0
Package: rog-control-center / asusctl (pkgbuilds/asusctl/PKGBUILD, pkgver 6.3.8)

## Problem
rog-control-center 6.3.8-1 aborts on startup, every launch, on a ROG Flow Z13
(GZ302EA_GZ302EA), Omarchy stable. Keyboard attached or not is irrelevant.

Systemd-coredump backtrace (two separate crashes, identical offsets):
  signal SIGABRT (SI_TKILL), main thread
  abort -> raise -> std::panicking -> bounds-check

Live reproduction:
  thread 'main' panicked at rog-control-center/src/ui/setup_aura.rs:148:55:
  index out of bounds: the len is 7 but the index is 255

## Root cause
setup_aura.rs:148 does `names[(*z) as usize]` — a raw power-zone value is used as
an index into the 7-entry power-zone name list. The Z13's N-KEY aura device
(xyz.ljones.Aura at /xyz/ljones/aura/18c6_4_5) reports SupportedPowerZones = [255]
(PowerZones::None; its aura_support.ron entry is power_zones: [r#None]), so the
index overflows. `find_aura_iface()` uses paths.pop() and lands on that interface.

## Fix exists upstream
Released tag 6.4.0 inserts `pow3r.retain(|z| *z != PowerZones::None)` before that
indexing (verified in 6.4.0 source). Tags 6.3.9–6.4.0 exist upstream.

## Request
Bump pkgver=6.3.8 -> 6.4.0 in pkgbuilds/asusctl/PKGBUILD and rebuild
asusctl + rog-control-center (or backport the one-line retain to 6.3.8).

Workaround until then: remove/adjust the GZ302 `product_id: "18c6"` entry in
/usr/share/asusd/aura_support.ron (power_zones: [r#None]) and restart asusd.

System: Omarchy stable, rog-control-center 6.3.8-1, hostname z13,
ASUS ROG Flow Z13 GZ302EA.

Filed by big-pickle via opencode.

jeremydixon22:
Closes #154.

This updates `asusctl` and `rog-control-center` from 6.3.8 to 6.4.0, follows the project source migration from GitLab to `opengamingcollective/asusctl` on GitHub, and refreshes the BLAKE2 checksum.

Because this is AUR-backed, the source/version change is also retained as an Omarchy patch so a later AUR sync cannot silently revert it. The checked-in package and a fresh raw-AUR-plus-overlay replay are byte-identical.

The version-bound overlay deliberately fails closed when the raw AUR recipe advances. That prevents this temporary upstream-ahead recipe from being silently transplanted onto a different release; once AUR catches up, the overlay should be reviewed and retired.

Verification:
- confirmed tag `6.4.0` resolves to upstream commit `e6c1469ccf2a745c6a1aff763852df90066c6baa`
- passed `makepkg --verifysource --holdver`
- built `asusctl`, `rog-control-center`, and the debug package successfully in `archlinux:base-devel`
- replayed the pinned AUR source through `bin/package-worktree`; `patched/` matches `current/`
- separately confirmed the overlay applies to the pinned raw recipe with `patch --dry-run --fuzz=0`
- passed `omarchy-pkgs self-test`, `omarchy-release self-test`, and the package-scoped edge release dry-run in the Arch environment

I could not reproduce the ROG Flow hardware startup path in the build environment, so the verification establishes the requested upstream version, source integrity, reproducible Omarchy overlay, and successful package build—not a hardware runtime claim.

### DIFF
--- a/pkgbuilds/asusctl/.omarchy/patches/6.4.0-github-source.patch
+++ b/pkgbuilds/asusctl/.omarchy/patches/6.4.0-github-source.patch
@@ -1,0 +1,18 @@
+diff --git a/PKGBUILD b/PKGBUILD
+index 55d0c22..8897aec 100644
+--- a/PKGBUILD
++++ b/PKGBUILD
+@@ -6,7 +6,7 @@ pkgname=(
+   asusctl
+   rog-control-center
+ )
+-pkgver=6.3.8
++pkgver=6.4.0
+ pkgrel=1
+ pkgdesc="A control daemon, tools, and a collection of crates for interacting with ASUS ROG laptops"
+ arch=('x86_64')
+@@ -28,2 +28,2 @@
+-source=("git+https://gitlab.com/asus-linux/asusctl.git#tag=$pkgver")
+-b2sums=('01269bc9fe63c1ce37568f9085c25dbcaaaa5cd9f51339d4c6904600d179ec826dace289d6e4e3501988fce5800da98e9cfa18b21c40da152d91c4f7fe821ffe')
++source=("git+https://github.com/opengamingcollective/asusctl.git#tag=$pkgver")
++b2sums=('0cfdc7f792fc6499cde9e4c9c6d82e484605e1b2878bb9f3a0e0696f5910b9ba7898f124fecd5fb3f39a0787f81e5163721606290099d757365dc3d6715a1bcf')
--- a/pkgbuilds/asusctl/PKGBUILD
+++ b/pkgbuilds/asusctl/PKGBUILD
@@ -6,8 +6,8 @@ pkgname=(
   asusctl
   rog-control-center
 )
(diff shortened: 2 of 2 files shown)


jeremydixon22:
The version-bound context is intentional fail-closed behavior in the current overlay path: it forces review when AUR advances instead of silently carrying a temporary upstream-ahead recipe onto a different release. I updated the PR description to make that retirement point explicit. A no-op `post-sync.sh` would still mark every future sync as customized and apply a local pkgrel suffix, while `sync: false` would unnecessarily disable AUR movement entirely.

PR #215, opened 2026-08-27
Reject fuzzy AUR overlay patches

jeremydixon22:
AUR overlays currently use GNU patch's default fuzz allowance. That can silently ignore changed context and transplant an Omarchy customization onto an upstream recipe it was not written against.

This moves patch application into one shared helper and sets `--fuzz=0` for both the live `sync-aur` path and the non-mutating `package-worktree` replay path. Exact context and ordinary line offsets remain supported; omitted or mismatched context fails closed for maintainer review.

The built-in self-test now contains two fixtures:
- an exact-context patch that must apply; and
- a patch that permissive GNU patch accepts with fuzz 1 but the shared helper must reject.

Verification:
- `bash -n` passes for the helper and all three changed scripts
- `omarchy-pkgs self-test` passes in `archlinux:base-devel`
- `v4l2-relayd` successfully replays all three existing overlays through the strict helper, with `patched/` matching `current/`

### DIFF
--- a/bin/omarchy-pkgs
+++ b/bin/omarchy-pkgs
@@ -637,10 +637,11 @@ cmd_self_test() {
   }
 
   check_strict_patch_application() {
-    local fixture_root exact_patch fuzzy_patch
+    local fixture_root exact_patch fuzzy_patch missing_root
     fixture_root=$(mktemp -d)
     exact_patch="$fixture_root/exact.patch"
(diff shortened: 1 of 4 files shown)


jeremydixon22:
Addressed the review note in the latest commit: the helper now explicitly exits before invoking patch if entering the package directory fails. Rechecked syntax, exact application, failed-directory behavior, and diff cleanliness.

PR #221, opened 2026-08-27
Add safe Windows firmware extraction for Snapdragon laptops

birkskyum:
## Summary

Adds `qcom-firmware-extract` as a fallback for firmware not already available to Linux. It reads the running device tree and stages missing files from a readable Windows driver store or backup directory (`-d`), so the installer can preserve them before partitioning.

- Installs under `/usr/lib/firmware/updates`. Differing variants require a unique companion-hash match within the same device-tree node; ambiguous or missing files are skipped without stopping other files.
- Matches the kernel's plain, zstd and xz firmware lookup. Gzip files do not hide missing loadable firmware.
- Includes board-selected GPU zap firmware in the initramfs and rebuilds when that configuration changes, even without extraction.

Discovery requires a device tree. An empty result on an ACPI-only machine does not establish firmware completeness. This does not download OEM packages or provide Linux drivers.

## Dependencies

Targets `omarchy-pkgs/master`. Consumers in omacom/omarchy#8672, omacom/omarchy#8673 and omacom/omarchy-iso#129 are merged into their respective `dragon` branches. This package can land and be published independently of #222 and #223.

Jimmy Van Veen authored the original extractor; Birk Skyum added matching and partial-install fixes. Original authorship is preserved.

## Validation

September 16 verification passed against master `5fe2367`: native ARM64 build of 1-4, extractor regressions, ShellCheck and whitespace checks. [Hosted self-tests and build-isolation checks pass](https://github.com/omacom/omarchy-pkgs/actions/runs/35038356179).

The current extraction/install path has not been rerun on the physical Yoga.

<details>
<summary>Earlier validation</summary>

The [September 6 integration build](https://github.com/birkskyum/omarchy-pkgs/actions/runs/34003828582) built the extractor and shim. Earlier Windows extraction started ADSP and CDSP on the Yoga, but CDSP startup alone does not demonstrate compute acceleration.

The September 10 image included this package and booted without extraction; its missing CDSP companion remained unresolved. [Integrated results and limits](https://github.com/omacom/omarchy-iso/pull/129) describe that image, not a new extraction test.

</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,9 +25,16 @@ jobs:
             -w /workspace \
             archlinux:base-devel bash -lc '
               set -euo pipefail
-              pacman -Syu --noconfirm git jq
+              pacman -Syu --noconfirm git jq python-yaml xz zstd
               ./bin/sync-upstream self-test
               ./bin/sync-rebuilds --self-test
               ./bin/omarchy-pkgs self-test
               ./bin/omarchy-release self-test
+              shopt -s nullglob
+              for test in pkgbuilds/*/test.sh; do
+                bash "$test"
+              done
+              if [[ -d test ]]; then
+                runuser -u nobody -- python -m unittest discover -s test -v
+              fi
             '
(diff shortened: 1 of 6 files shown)


JimmayVV:
Confirming provenance. `qcom-firmware-extract` is mine. The companion-hash matching is a real improvement. The "newest wins" rule was only ever tested against one driver store. Thanks for catching it.

PR #222, opened 2026-08-27
Add linux-aarch64 kernel compatibility shim

birkskyum:
## Summary

Adds `linux-aarch64-pkgbase-shim` so mkinitcpio and Limine can use Arch Linux ARM kernels shipped as `/boot/Image`.

- Supplies module-directory `pkgbase` and `vmlinuz`, refreshing the image on kernel upgrades and changed same-version reinstalls.
- Preserves package-owned files, copies atomically with original permissions, and removes stale shim-only metadata without deleting DKMS leftovers.
- Defers regeneration when the installer masks the normal hook and reports rebuild failures.

The shim remains needed until the kernel package provides the discoverable image directly.

## Dependencies

Targets `omarchy-pkgs/master`. The installer in omacom/omarchy-iso#129 is merged into `dragon` and installs this package. It can land and be published independently of #221 and #223.

Jim Martin's #380 addresses a separate update-safety problem: retaining encryption and UKI configuration. The shim does not replace that fix.

Jimmy Van Veen authored the original shim; original authorship is preserved.

## Validation

September 16 verification passed against master `5fe2367`: native ARM64 build of 1-3, shim regressions, ShellCheck and whitespace checks. [Hosted self-tests and build-isolation checks pass](https://github.com/omacom/omarchy-pkgs/actions/runs/35038354164).

An isolated ARM container tested a real 7.2.3-to-7.2.4 kernel upgrade and same-version reinstall alongside a local #380 refresh for settings 4.0.4. Metadata was refreshed before the rebuild hook; real mkinitcpio output retained encryption and Plymouth support, and settings preserved a local edit.

This used a Limine backend fixture. Physical update/reboot, ESP deployment and Asahi boot remain untested.

<details>
<summary>Earlier validation</summary>

The [September 6 integration build](https://github.com/birkskyum/omarchy-pkgs/actions/runs/34003828582) built the shim and extractor. September 10 container tests also covered kernel upgrade/reinstall, stale-metadata cleanup and encrypted-root initramfs generation alongside #380.

The shim was included in the September 10 image that booted on the Yoga. That did not test upgrading the installed system. [Integrated results and limits](https://github.com/omacom/omarchy-iso/pull/129) retain that distinction.

</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,9 +25,16 @@ jobs:
             -w /workspace \
             archlinux:base-devel bash -lc '
               set -euo pipefail
-              pacman -Syu --noconfirm git jq
+              pacman -Syu --noconfirm git jq python-yaml xz zstd
               ./bin/sync-upstream self-test
               ./bin/sync-rebuilds --self-test
               ./bin/omarchy-pkgs self-test
               ./bin/omarchy-release self-test
+              shopt -s nullglob
+              for test in pkgbuilds/*/test.sh; do
+                bash "$test"
+              done
+              if [[ -d test ]]; then
+                runuser -u nobody -- python -m unittest discover -s test -v
+              fi
             '
--- a/pkgbuilds/linux-aarch64-pkgbase-shim/.omarchy/package.json
+++ b/pkgbuilds/linux-aarch64-pkgbase-shim/.omarchy/package.json
@@ -1,0 +1,3 @@
+{
+  "source": "local"
+}
--- a/pkgbuilds/linux-aarch64-pkgbase-shim/85-linux-aarch64-pkgbase-shim.hook
+++ b/pkgbuilds/linux-aarch64-pkgbase-shim/85-linux-aarch64-pkgbase-shim.hook
@@ -1,0 +1,20 @@
+# Ordered before 90-mkinitcpio-install (limine-mkinitcpio-hook's override of
+# it included) so the pkgbase and vmlinuz it writes are there when that hook
+# globs usr/lib/modules/*/pkgbase.
+[Trigger]
+Type = Path
+Operation = Install
+Operation = Upgrade
+Target = usr/lib/modules/*/
+
+[Trigger]
+Type = Package
+Operation = Install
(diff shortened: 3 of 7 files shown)


birkskyum:
Native ARM64 validation: the cleaned package stack built successfully in the combined PR integration run. The run failed only in the later repository-database publishing step owned by #223, after all three requested packages had been produced. Build evidence: https://github.com/birkskyum/omarchy-pkgs/actions/runs/33118509188

JimmayVV:
I think one piece of the shim this branch retires is still needed. Zesko !64 lets `limine-mkinitcpio-install` discover a kernel whose package ships no `pkgbase`. But the hook then runs `mkinitcpio --kernel <version>` without `--kernelimage`, and mkinitcpio only looks for the image at `/lib/modules/<ver>/vmlinuz`, `/lib/modules/<ver>/vmlinux` and `/boot/vmlinuz-*`. Arch Linux ARM installs `/boot/Image`, which matches none of those. So I'd expect the UKI build to fail with "Could not find kernel image" once nothing copies `Image` into the modules directory. That copy was the other half of what `linux-aarch64-pkgbase-shim` did, alongside writing `pkgbase`.

If your Yoga test ran with the shim still installed from the integration branch, that would explain it working. Two fixes I can see. Keep the shim here with its retirement note, or have this patch also point mkinitcpio at `/boot/Image`. Happy to be wrong if I've missed something in the branch.

birkskyum:
Good catch. Confirmed: MR 64 discovers the module directory but mkinitcpio still cannot find Arch Linux ARM's /boot/Image. I restored your three shim commits with authorship intact and added an ownership compatibility patch after MR 64, because the shim-written pkgbase is not package-owned. The PR is back in draft until the corrected exact stack passes a native ARM build.

birkskyum:
Corrected native ARM validation passed, including linux-aarch64-pkgbase-shim 1-3, limine-mkinitcpio-hook 1.37.1-4, repository database creation, verification, and artifact upload: https://github.com/birkskyum/omarchy-pkgs/actions/runs/33122253609

PR #227, opened 2026-08-28
Require inotify-tools for snapshot watching

jeremydixon22:
## Summary

- move `inotify-tools` from optional to required dependencies for `limine-snapper-sync`
- preserve the correction as an Omarchy overlay across future AUR syncs
- publish the customized build as `1.31.0-1.1`

The watcher explicitly needs `inotifywait`; without it, the service exits and falls back instead of watching snapshot changes. This addresses the package-dependency portion of basecamp/omarchy#6629. Creating a Snapper root configuration remains a separate system-policy concern.

## Verification

- exact AUR commit `35aff513bf1f0284380c8592d3de213f00bb6a7b`
- overlay applies with `--fuzz=0`
- `bin/sync-aur limine-snapper-sync` is idempotent
- emitted `.SRCINFO` lists `inotify-tools` under `depends`, not `optdepends`
- `makepkg --verifysource` passed for the Git source and x86_64 GraalVM archive

### DIFF
--- a/pkgbuilds/limine-snapper-sync/.omarchy/patches/require-inotify-tools.patch
+++ b/pkgbuilds/limine-snapper-sync/.omarchy/patches/require-inotify-tools.patch
@@ -1,0 +1,9 @@
+diff --git a/PKGBUILD b/PKGBUILD
+--- a/PKGBUILD
++++ b/PKGBUILD
+@@ -19 +19,2 @@
+-	'libnotify')
++	'libnotify'
++	'inotify-tools')
+@@ -23 +24,0 @@
+-	'inotify-tools: Monitors when snapshots are created or deleted.'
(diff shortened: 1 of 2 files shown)


ryanrhughes:
Closing this because `limine-snapper-sync` is AUR-synced and this would introduce a permanent downstream overlay without an Omarchy-specific requirement.

Upstream intentionally made `inotify-tools` optional and ships the Snapper plugin as the supported fallback. When `inotifywait` is absent, the watcher exits successfully, so `inactive (dead)` is expected in that mode. Requiring `inotify-tools` also would not fix the separately reported missing root Snapper configuration.

If continuous inotify monitoring is genuinely required by Omarchy, that requirement and why the supported plugin path is insufficient should be demonstrated first; otherwise the packaging policy change belongs upstream/AUR. Thanks for the contribution.

PR #248, opened 2026-08-31
gnome-keyring: escape multiline secrets in textual keyrings

Irfrit:
Backports GNOME MR [GNOME/gnome-keyring!111](https://gitlab.gnome.org/GNOME/gnome-keyring/-/merge_requests/111)to fix secret corruption when storing multiline secrets (e.g. JSON certificates) in textual keyrings.


Irfrit:
@ryanrhughes fixed the build dependencied. Now it passes

PR #254, opened 2026-09-01
Add Pixygon Linux apps (reel, infinite)

gentlemanSoup:
Prebuilt `-bin` packages for Pixygon's Linux-native apps.

Each PKGBUILD is generated from the project's public release manifest
(`https://pixygontech.b-cdn.net/releases/<app>/manifest.json`), so `pkgver` and `sha256sums` always
match the published artifact — refreshed automatically whenever we ship.

- **pixygon-infinite-bin** 0.127.0
- **reel-bin** 1.14.0


Chessing234 (changes requested):
all three pkgbuilds hardcode release path versions while pkgver is parameterized — next bump will fetch the wrong tarball. use v${pkgver} (or equivalent) in the download urls.

PR #257, opened 2026-09-01
[codex] Revoke passwordless grants when settings is removed

AFOliveira:
Settings owns the complete temporary sudo-grant lifecycle: publication helper, shared security library, boot cleanup and an aborting pre-transaction revocation hook. Removing only the desktop runtime preserves expiry support; upgrading or removing settings revokes the grant before those files disappear.

The hook and grant publication share one lock and a package-transition marker. Revocation failure aborts the ALPM transaction and blocks new grants until a successful retry. Scriptlets retain the legacy fallback for the first upgrade from packages without the hook; scriptlet errors are not claimed to abort that transaction.

Companion: [Omarchy #9457](https://github.com/omacom/omarchy/pull/9457). Rebased onto current `master` (`v4.0.4`, `c668141e`) as three commits with no release bump of their own: the runtime packages now require the exact `pkgver-pkgrel` of their matching settings package in both channels, settings gains the `sudo` dependency, the hook and the two helpers. #239 is separate overlapping ownership work, not a required parent.

**Not mergeable before #9457: the declared upstream source `c668141e` predates it and cannot build these settings recipes.** After #9457 lands, advance both release pins and checksums together to the real upstream commit containing it, bump `pkgrel` on the pair, build the declared sources, and publish each matching pair together. No fork revision or duplicate backport has been substituted for that upstream prerequisite.

Validation at the rebased head: all four PKGBUILDs parse (`makepkg --printsrcinfo`); the cross-repository suites in #9457 at `ea617b91` (`passwordless-grant-lifecycle-test.sh`, `nopasswd-sudo-expiry-test.sh`) pass with `OMARCHY_PKGS_PATH` pointing at this checkout, including both settings scriptlets' revoke, block and recover paths and the native lock serialization against grant publication. Earlier heads were reviewed independently by Daybreak and exercised in a disposable VM (reinstall, stable/dev transfers, runtime and settings removal, aborting removal on a benign revocation error, recovery); native ARM runtime and fresh-ISO validation remain untested.


ErikMelton:
After changes:

  - [P1] Exact dev dependency breaks normal builds — PKGBUILD:21
    https://github.com/omacom/omarchy-pkgs/blob/32c1415490a5771c264ab0bfa854f50b14d0de06/pkgbuilds/omarchy-dev/PKGBUILD#L21  makepkg checks dependencies before running pkgver(), so it requests the stale omarchy-settings-dev=4.0.0.r847.g4de185b-1 even after the builder produces newer settings. The normal build path skips its preliminary version refresh for integer pkgrels. I reproduced this dependency failure. Refresh the dev version before dependency resolution while retaining the exact dependency in the resulting package.

AFOliveira:
Fixed in 21b3b32, now at 9b9a9c8e. The builder's VCS refresh runs for every VCS package, so dependency resolution sees the refreshed pkgver and the package keeps the exact settings dependency; the real build uses `--holdver` so it packages the revision it validated. The other integer-pkgrel VCS packages are all commit-pinned, so only the redundant fetch goes away.

The same head hardens the scriptlets: completion writes the blocker before waiting for the lock, sweeps the reserved prefix, and clears the blocker only when nothing remains and the boot cleanup is present, so a failed `pre_upgrade` followed by `post_upgrade`, an interrupted removal, or a fresh install with a leftover rule all leave publication refused. Recovery runs before the `/etc` overrides, lock waits are bounded to 60s, and inherited glob state is cleared. The cross-repo contract test in #9457 covers each case. Still not mergeable before #9457 lands and the pin advances.

AFOliveira:
@ryanrhughes this is not a feature, but a security fix.

PR #263, opened 2026-09-01
Disable YT6801 private ioctl interface

AFOliveira:
Removes the YT6801 vendor driver's private ioctl callbacks, linked object and dormant implementation from the shipped DKMS source. This protects systems still using the external driver while [Omarchy #9682](https://github.com/omacom/omarchy/pull/9682) moves supported systems to the in-tree driver; neither change replaces the other.

The package uses the official Motorcomm 1.0.34 archive with mandatory checksums, applies the patch without fuzz, and rejects missing/unreadable verification inputs or grep errors. Release `1.0.34-2` includes these checks. Automated AUR replacement remains disabled so it cannot discard the security patch.

Validation at `3015e9f`: benign negative checks and syntax/whitespace passed; an actual declared-source `makepkg` build passed; the exact packaged sources compiled against signature-verified Arch `7.1.9-arch1-2` headers. The archive omits the private implementation, and the built module has no private ioctl handlers or object. Independent Daybreak review and follow-up found no remaining code or package-build blocker. Builds ran unprivileged with dependency checks skipped and without installation or module loading.

The canonical vendor URL is mutable. A changed download fails checksum validation; no immutable official alternative was found. Other kernel versions and real NIC traffic, suspend/resume and reboot remain untested.


kwilczynski:
Looks good!

The vendor might fix some of these issues in the next release. However, I wouldn't personally count on it, thus removing this private ioctl interface, which currently is only used by the troubleshooting and debugging tools, would be fine here. :+1:

kwilczynski (approved):

PR #264, opened 2026-09-01
Harden IPU7 PSYS userptr pinning

AFOliveira:
Hardens IPU7 PSYS userptr pinning and its failure cleanup. The driver rejects invalid/oversized ranges, uses overflow-checked page-array allocation, removes forced write access, unwinds partial pins safely, and clears attachment state after release. It also clears `db_attach` on attachment/map failure before an exported buffer can survive for later fd release; Daybreak found that adjacent stale-pointer defect during this audit.

The package replaces world-writable device/runtime permissions with video/seat-scoped access. Both production driver patches apply with zero fuzz. The package check propagates source/tool failures and preserves a missing-module failure through cleanup. Release advances to `1.0.5-3`.

Validation at `4d00f12`: exact pinned Intel source accepts both patches with zero fuzz; local source checksums match; strict Linux checkpatch reports zero errors/warnings; the range helper passes. All six patched driver modules compile against Arch `7.1.9-arch1-2` headers. Benign fixtures confirm missing input, grep errors and a successful build command without its required module all fail the package check. The final prose-only patch metadata change produces the identical tested C blob. Daybreak's final review accepts the fix.

The complete proprietary HAL/package payload, other kernels and actual IPU7 hardware were not validated in this audit. #271 is closed without merging and does not establish kernel compatibility. Repeated valid long-term mappings and aggregate pin quotas remain a separate availability issue. If the old module is loaded, a reboot is required to replace it. Permission refresh is attempted during upgrade; failures retain the existing reboot-before-use warning and do not skip unrelated package lifecycle work.


ryanrhughes:
@spencerbull can you take a look at this one?

PR #274, opened 2026-09-02
Add claude-desktop, tracked from Anthropic's signed apt index

evindor:
## What

Adds `claude-desktop`, Anthropic's official Claude desktop app (Chat, Cowork, and Claude Code), tracked from Anthropic's own Debian repository.

Anthropic ships the Linux app only as a `.deb` from their apt repository, and their docs send Arch users to the CLI instead. This package unpacks that `.deb` and nothing else — no patching, no rebuilding, no bundled extras.

## Why not the AUR

Same reasoning as `openai-codex-desktop`: releases land several times a week (six in the last three weeks), and tracking the vendor's index directly keeps up without waiting on an AUR maintainer.

There is a second reason here. **Anthropic signs their apt index**, so this package can do something the AUR route can't: refuse to accept a checksum that Anthropic hasn't signed.

## How the checksums are trusted

`.omarchy/upstream.sh` walks a chain that starts at a key committed in this repo:

```
bundled key, pinned by fingerprint  ->  signs InRelease
InRelease                           ->  hashes Packages
Packages                            ->  hashes each .deb
```

Every checksum that reaches the PKGBUILD arrives under Anthropic's own signature. A break anywhere fails the sync instead of proposing a checksum nobody vouched for. Verified failing closed against: a fingerprint that doesn't match the anchor, a substituted key, a missing key, and an attacker-modified `Packages` served under a genuine `InRelease`.

The key is committed rather than fetched, so rotating it is a reviewed change to this package rather than something the server can do to us. Its fingerprint `31DDDE24DDFAB679F42D7BD2BAA929FF1A7ECACE` was cross-checked against four independent sources: the published install docs, `downloads.claude.ai/claude-desktop/key.asc`, the `InRelease` signature itself, and the copy embedded in the `.deb`'s own `postinst`.

`min_release_age: "24h"` is set, so an automated sync will not ship a release that upstream hasn't had a day to pull. The hook reports `published_at` from the pool's `Last-Modified`, and falls back to the newest release that has already cleared the window rather than stalling on a fresh one.

## What is dropped from the .deb

The `.deb`'s maintainer scripts are not reproduced. They do three Debian-specific things, none of which applies here:

1. register Anthropic's apt repository and an unattended-upgrades snippet,
2. install an AppArmor profile gated on Ubuntu 24.04's userns restriction,
3. register a GNOME Shell search provider.

The payload itself is confined to `/usr` and carries one setuid binary, `chrome-sandbox` — Chromium's sandbox helper, shipped 4755 the way every Chromium-based package does.

## Keyring on Hyprland

Chromium cannot reliably infer the Secret Service password-store backend from a Hyprland session even when GNOME Keyring is already providing it, so the app intermittently decides no keyring is available, refuses to persist the sign-in, and shows "Your sign-in won't be saved on this device". This is the same thing `hermes-desktop` hit, and 246eea9 is the same fix — by flag rather than env var, because that is what this app reads.

The `.deb` ships `/usr/bin/claude-desktop` as a symlink to the Electron binary; this replaces it with a launcher that names `gnome-libsecret`. Three cases are left alone: an explicit `--password-store` from the caller, KDE (the app appends `--password-store=kwalletd6` itself and carries fallback logic for a KWallet with no wallet), and `CLAUDE_DESKTOP_PASSWORD_STORE=none`. `libsecret` is already a dependency.

I hit this on a fresh install: the toast appeared, yet the keyring was healthy the whole time — `gnome-keyring` running, owning `org.freedesktop.secrets`, collection unlocked, and a `Claude Safe Storage` item created one second after first launch. "Not reliably" is exactly right; it is a race, not an absent keyring.

## Cowork

Cowork boots its agent in a QEMU VM. Upstream ships that stack under `Recommends:`, which apt installs by default, so it is a hard dependency here to give the same working-out-of-the-box Cowork that Debian and Ubuntu users get. Happy to move `qemu-system-x86`/`edk2-ovmf` to `optdepends` if you'd rather not pull a virtualisation stack onto every machine — `virtiofsd` does need to stay a real dependency, since the app resolves it from the system before its own bundled copy.

Two symlink shims map Debian's firmware and `virtiofsd` paths onto Arch's. aarch64 needs no firmware shim; `edk2-aarch64` already installs at the path the app opens.

## Testing

- `bin/sync-upstream self-test` passes.
- `bin/sync-upstream claude-desktop` reports correctly against the checked-in pin, and rewrites `pkgver`, `pkgrel` and both checksum arrays correctly when the pin is behind.
- Builds clean for x86_64 on Omarchy 4.0.2; the resulting package contains only `/usr`, the expected setuid `chrome-sandbox`, the two shims, and the launcher.
- The launcher's branches were exercised against a stub: default, URL passthrough, both `--password-store` spellings, `CLAUDE_DESKTOP_PASSWORD_STORE` set and `none`, KDE and Plasma, and an unset `XDG_CURRENT_DESKTOP`.
- The pinned `1.40609.1` checksums were verified by hand against the signed index and against a downloaded `.deb`.

I don't have aarch64 hardware, so that build is unverified beyond the checksums coming from the signed index.

## Follow-up, if this merges

This only makes the package installable (`omarchy pkg add claude-desktop`). It does not add a menu entry — Install → AI lists ChatGPT Desktop, Dictation, Grok Bot, LM Studio and Ollama, and Claude would still be missing from it.

That belongs in `omacom/omarchy` rather than here, and needs no new script — the Grok Bot / LM Studio shape covers it, since unlike ChatGPT there is nothing to launch through `uwsm-app`:

```jsonc
"install.ai.claude": {"icon":"󰛄","label":"Claude Desktop","when":"! omarchy-pkg-present claude-desktop","action":"omarchy-install-and-launch 'Claude Desktop' claude-desktop com.anthropic.Claude"},
```

`com.anthropic.Claude` is the desktop-id the package installs, and `󰛄` is the glyph the default-agent menu already uses for Claude. It has to land after this one, since both the `when` guard and the action name a package that does not exist yet. Happy to open that separately once this is in, or to drop it if you would rather Claude Desktop stay install-by-name only.


ryanrhughes:
Superseded: claude-desktop is on master (now 2.2553.1) tracked from the same signed apt feed.

Issue #199, opened 2026-08-24
Publish an aarch64 tree at pkgs.omarchy.org


oceanapplications:
`pkgs.omarchy.org` serves only x86_64 today:

| | `stable/$arch/omarchy.db` | `edge/$arch/omarchy.db` |
|---|---|---|
| `x86_64` | 200 | 200 |
| `aarch64` | **404** | **404** |

`[omarchy]` is defined as `Server = https://pkgs.omarchy.org/stable/$arch`, and pacman
substitutes `$arch` from `Architecture` in `pacman.conf`. So on ARM64 that repository simply
does not resolve, and an installed Omarchy cannot install or update any Omarchy package from
the official source.

### The tooling already supports it

This is a publishing step rather than a code change — `bin/build` already takes the flag:

```
--arch <arch>        Target architecture (x86_64 or aarch64, default: x86_64)
```

along with `update_arch_paths` and QEMU setup for building aarch64 on an x86_64 host, and
`bin/sync-repo` takes the same `--arch`. The README documents the QEMU prerequisite under
**aarch64 Builds**. As far as I can tell the only missing piece is running that path for
aarch64 and syncing the result.

### Why it matters now

aarch64 Omarchy works. It runs on Arch Linux ARM under Parallels on Apple Silicon —
GPU-accelerated via virgl, Hyprland under uwsm, Limine with UKI and snapper rollback — and an
aarch64 ISO builds, installs to an encrypted target, and boots unattended into the desktop.

Everything needed to *build* the packages for aarch64 is either already here or proposed:

- #195 — nine packages that build natively on ARM64 gain `aarch64` in `arch=()`
- #197 — `1password`, which upstream publishes an aarch64 build of

What is missing is somewhere to *put* them. Without a published tree, anyone wanting to run
Omarchy on ARM64 has to build roughly 34 packages locally and serve them from a `file://`
repository, which is what I have been doing. That makes the work unreproducible for anyone
else, and it is the only remaining blocker of its kind — every other gap I hit has a patch
proposed against it:

| | |
|---|---|
| omacom-io/omarchy-iso#121 | aarch64 ISO build |
| basecamp/omarchy#8039 | three x86-only assumptions in the install path |
| Zesko/limine-entry-tool!63 | Limine's unnecessary x86_64 gates (merged) |
| Zesko/limine-entry-tool!64 | Limine finding kernels that ship no `pkgbase` marker (merged, awaiting a release) |

### What I am asking for

[...]

PR #277, opened 2026-09-02
Make aarch64 a first-class architecture in the scheduled pipeline

maralcbr:
## Summary

Make aarch64 a first-class architecture in the scheduled pipeline on the repository host, so enabling it is one line and the existing timers do the rest. Third of the Apple Silicon series (#275, #276). Nothing changes until an architecture is added to the published list.

### One list drives the pipeline

`helpers/paths.sh` gains `PUBLISHED_ARCHES` (default `x86_64`, overridable with `OMARCHY_ARCHES`). Everything scheduled runs once per entry:

- **`check-versions`** compares PKGBUILDs against each architecture's channel databases and writes one queue per channel and architecture, `.sync-needed-<channel>-<arch>`. It also accepts `--arch` for a single one.
- **`auto-release <channel> [<arch>|all]`** works through the queues one architecture at a time, each with its own backoff file, so a failing aarch64 build never blocks x86_64 or the other way round. The timers keep calling `auto-release <channel>` unchanged; `all` is the default. A queue or backoff file written under the old per-channel name is treated as x86_64 until it is consumed, so a host upgraded mid-cycle loses nothing.
- **`advance-channel --arch all`** re-runs the advance for every published architecture, stopping at the first failure. `omarchy-release` uses it for `start` and `ship`, so a train never moves a channel for one architecture and not another, and its rc trigger builds the pinned pair once per published architecture.
- **`omarchy-release`** observes channels through the reference architecture (the first in the list) instead of a hard-coded `x86_64`.
- **`timers`** shows queues and backoff per architecture. **`setup`** prints the published list.

### The two things that made an aarch64 image impossible to build

- **Keyring bootstrap.** The builder image installed `omarchy-keyring` from `pkgs.omarchy.org/<channel>/$arch`, which for aarch64 does not exist yet (#199), so the image that would create that tree could not be built. `omarchy-keyring` is `arch=any`, so the image now always takes it from the x86_64 tree, whatever its own architecture.
- **QEMU probe.** `bin/build` only set up emulation for the x86_64-host, aarch64-target case. It now compares the host and target architectures and probes by running a container for the target platform, so an aarch64 host building x86_64 works the same way.

### Smaller fixes

- `clean-repo` grouped versions with a filename regex that only knew `any|x86_64|i686`, so aarch64 packages would never have been pruned. The architecture is now whatever sits between the last hyphen and `.pkg.tar`.
- README: how the published list works, how to enable an architecture, and that Arch Linux ARM offers no dated snapshots to pin a channel's base to (so aarch64 uses one mirrorlist for all three channels; worth a decision if per-channel pinning matters for the fast ring).

### Not in this PR

- `sync-rebuilds` still compares against x86_64 only. Its `rebuilt_against` record has to become per-architecture before rebuild triggers can be trusted on a second one, as the README already says; that is its own change.
- Flipping `PUBLISHED_ARCHES` to include aarch64. That is the enablement step and belongs to the maintainer, once the channel tree is ready to be published.

## Testing

- Arch Linux ARM container with a stubbed `bin/repo`, two published architectures, empty channel databases: `check-versions` writes six queue files; `auto-release edge` releases x86_64, fails aarch64 and backs it off without touching x86_64; the next tick reports the backoff and does nothing else; a legacy `.sync-needed-edge` is picked up as x86_64 and consumed; `advance-channel --from edge --to rc --arch all --dry-run` runs the advance for both. `omarchy-pkgs` and `omarchy-release` self-tests pass.
- MacBook Pro M1 Pro with native Docker, unpatched Dockerfile: `bin/build --arch aarch64 --package retroarch-joypad-autoconfig-git` builds the aarch64 edge builder image (keyring from the x86_64 tree) and the package. Before this change the same command failed in the image build with a 404 on `edge/aarch64/omarchy.db`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
               set -euo pipefail
               pacman -Syu --noconfirm jq
               ./bin/sync-upstream self-test
+              ./bin/sync-rebuilds --self-test
               ./bin/omarchy-pkgs self-test
               ./bin/omarchy-release self-test
             '
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.firecrawl/
 src/
 logs/
 build-output/
--- a/README.md
+++ b/README.md
@@ -57,14 +57,25 @@ file per channel and architecture (`.sync-needed-<channel>-<arch>`);
 `auto-release <channel>` works through the queues one architecture at a
 time, each with its own backoff (`.build-failed-<channel>-<arch>`), so a
 failing build on one architecture never holds up the other; and the release
-train advances channels with `--arch all`, so a channel never moves for one
-architecture and not another. The first entry is the reference architecture
-the release train observes channels through.
+train advances channels with `--arch all`: it takes one host-wide lock and
+verifies every architecture's source database before moving any of them. The
+first entry is the reference architecture the release train observes channels
+through. A remote sync failure can still leave a promotion temporarily partial;
+rerunning the same advance completes it safely.
 
-Adding an architecture is therefore one change to that list (or
-`OMARCHY_ARCHES` in the host's build credentials): the next `check-versions`
-tick queues everything the new architecture lacks, and the next
-`auto-release` tick starts building it. The builder image bootstraps
+Adding an architecture to the scheduled pipeline is therefore one checked-in
+change to that list: the next `check-versions` tick queues everything the new
+architecture lacks, and the next `auto-release` tick starts building it. A
+checked-in list also means the rebuild workflow and release host cannot drift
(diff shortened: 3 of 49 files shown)


ryanrhughes (approved):
QA complete: both engine dry runs pass; rootless Podman built omarchy-keyring with host-owned 0644 outputs; Docker argument parity, engine selection/override, QEMU registration handling, shell syntax, ShellCheck, repository self-tests, and CI all pass.

ryanrhughes (approved):
Re-reviewed at ca99c24 after the force-update. Local x86_64/aarch64 and Docker/Podman QA passed, published ARM artifacts were verified, and GitHub self-tests pass on this exact head.

PR #284, opened 2026-09-03
Report Omarchy to Steam instead of Arch Linux

spaceXrace:
## Summary

- ship an Omarchy-specific `/etc/lsb-release` override
- apply it from both stable and development `omarchy-settings` packages
- keep the release value as `rolling` so reporting is not split across Omarchy patch versions

## Why

Omarchy already identifies itself correctly in `/etc/os-release`:

```text
ID=omarchy
ID_LIKE=arch
NAME="Omarchy"
```

However, Arch's `steam` package depends on `lsb-release`, whose `/etc/lsb-release` currently says:

```text
DISTRIB_ID="Arch"
DISTRIB_RELEASE="rolling"
DISTRIB_DESCRIPTION="Arch Linux"
```

Steam's launcher checks `/etc/lsb-release` before `/etc/os-release`. On a stock Omarchy 4.0.2 installation this is visible in `~/.local/share/Steam/logs/console-linux.txt`:

```text
Running Steam on arch rolling 64-bit
```

Steam's public hardware survey does not currently list Omarchy as a distinct Linux version. Although the public table only exposes entries above its reporting cutoff, the client-side value shows that native Steam installations submit/use the Arch identity at this detection point.

With this override, the same Steam detection path resolves to:

```text
omarchy rolling 64-bit
```

## Verification

- `bash -n` passes for both PKGBUILDs and install scripts
- all repository self-tests pass:
  - `./bin/sync-upstream self-test`
  - `./bin/omarchy-pkgs self-test`
  - `./bin/omarchy-release self-test`
- `omarchy-settings` builds successfully with `makepkg`
- inspected the built archive to verify the new override and install-script copy are present

### DIFF
--- a/pkgbuilds/omarchy-settings-dev/PKGBUILD
+++ b/pkgbuilds/omarchy-settings-dev/PKGBUILD
@@ -37,11 +37,11 @@ makedepends=(
 
 # backup=() is pacman's hook for protecting user edits to package-owned files
 # on upgrade. Only the paths the package actually installs into /etc qualify.
-# The etc-overrides set (faillock.conf, nsswitch.conf, cups-browsed.conf,
+# The etc-overrides set (os-release, lsb-release, faillock.conf, nsswitch.conf, cups-browsed.conf,
 # cups-files.conf, plymouthd.conf, /etc/skel/.bashrc) ships at
 # /usr/share/omarchy/etc-overrides/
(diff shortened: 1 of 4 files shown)


ryanrhughes:
Superseded: the Steam os-release change is on master in omarchy-settings.

PR #304, opened 2026-09-04
Add omsticker USB creator package

paulogeyer:
Adds **OmSticker**, a Qt6 app that writes a bootable ISO to a USB stick and formats leftover space (exFAT, FAT32, NTFS, ext4, Btrfs, F2FS, or XFS). It can flash any ISO; downloading the latest Omarchy image is the default convenience.

- Source: https://github.com/paulogeyer/omsticker (MIT)
- Release: https://github.com/paulogeyer/omsticker/releases/tag/v0.9.1
- Packaging: local package, checksum-verified GitHub tag tarball
- Installs the binary, helper, desktop entry, and icon
- Flashing uses UDisks2 (no custom polkit policy)

Checks: PKGBUILD against the v0.9.1 archive; `sha256sums` matches `v0.9.1.tar.gz`.

Happy to drop or retarget the ring if you prefer it only on edge.


Chessing234 (approved):
qt build from tagged source with sha256.

PR #306, opened 2026-09-04
Add cua-driver-bin, the Cua computer-use driver, to the fast ring

spencerbull:
Adds [Cua Driver](https://github.com/trycua/cua) as a vendor-tracked binary package for x86_64 and aarch64 in the fast ring, with a 24-hour minimum release age. The monorepo's release feed interleaves several products and marks Driver releases as prereleases, so a package hook scans the paginated feed and selects the highest eligible version with a stable Driver tag. It validates publication dates and both architecture checksums, rejects incomplete scans, and handles large release pages through stdin.

The vendor payload stays together under `/usr/lib/cua-driver`, exposed through `/usr/bin/cua-driver`, because the CLI resolves its cursor-theme compiler relative to its real executable. The SDK library, Node module, ABI header and GNOME helper extension ship alongside it. This package does not include the optional Hyprland plugin proposed in #346.

The packaged 0.23.2 binary's vendor updater would otherwise install another copy under the user's home directory, bypassing pacman and the release-age gate. `prepare()` replaces both embedded installer URLs with an equal-length URL for the packaged `pm.sh` refusal stub. Occurrence-count and file-size checks stop the build if the expected rewrite no longer applies. The stub directs updates to pacman; it also covers configurations that invoke the resolved binary directly. Keep the workaround until a release containing trycua/cua#3636 reaches this package through its normal update path.

The existing repository self-test entry point runs deterministic release-hook fixtures against a temporary pinned PKGBUILD, keeping fixture versions independent of future package updates.

🤖 Generated by GPT-6 in T3 Code. Reviewed by Codex GPT-6 XHigh.

### DIFF
--- a/bin/sync-upstream
+++ b/bin/sync-upstream
@@ -834,6 +834,10 @@ EOF
   check "malformed upstream fails the sync instead of skipping" "1" "$((FAILED - prev_failed))"
   FAILED=0
 
+  if ! bash "$BUILD_ROOT/pkgbuilds/cua-driver-bin/.omarchy/upstream-test.sh"; then
+    failures=$((failures + 1))
+  fi
+
   echo ""
   if [[ "$failures" -eq 0 ]]; then
     print_success "Self-test passed"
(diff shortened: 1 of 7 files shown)


Chessing234 (approved):
nice work blocking the vendor curl|bash self-updater via binary url rewrite + pm.sh stub. checksums look solid.

PR #310, opened 2026-09-05
Add agent-arcade-bin to the fast ring

DanWahlin:
Adds **agent-arcade-bin**, a collection of six original retro arcade games (Alien Onslaught, Cosmic Rocks, Galaxy Blaster, Ninja Runner, Planet Guardian, and Surface Defense) that runs as a transparent always-on-top overlay while an AI coding agent is busy. Tauri + Phaser, MIT licensed, public source, x86_64 only.

- Repo: https://github.com/DanWahlin/agent-arcade
- Site: https://danwahlin.github.io/agent-arcade
- Release: https://github.com/DanWahlin/agent-arcade/releases/tag/v0.6.2

### Packaging and updates

Upstream publishes a `.deb` built on Ubuntu 22.04 via `tauri-action`, so this package repackages its `usr/` tree. `.omarchy/package.json` follows stable GitHub releases, reads GitHub's SHA-256 asset digest, uses the fast ring, and applies a 24-hour release quarantine.

The initial `0.6.2` pin was taken inside that window deliberately because it contains the package-manager integration required by this submission. Unattended syncs still honor the 24-hour floor; `0.6.2` clears it at `2026-09-15T06:53:43Z`.

Agent Arcade `v0.6.2` supports `/usr/share/agent-arcade/package-manager` as a Linux package-manager marker. This package installs that marker, which prevents the application from checking GitHub for updates or invoking Tauri's direct installer. Updates therefore remain controlled by pacman and the Omarchy release policy even when the packaged binary is launched directly.

### Dependencies

Dependencies were checked against the released binary. It links WebKitGTK 4.1 and GTK3; the tray icon uses `libayatana-appindicator`. Game audio uses WebKitGTK's GStreamer backend, so the base, good, bad, and libav plugin sets are runtime dependencies.

### Validation

- GitHub release workflow passed Linux, Windows, macOS, and release publication jobs.
- Verified the `v0.6.2` `.deb` SHA-256 against GitHub's published digest.
- Confirmed the released binary contains the package-manager marker check.
- Rebuilt `agent-arcade-bin 0.6.2-1` and confirmed the marker is packaged.
- Installed through pacman on Omarchy; `pacman -Qkk` reported 22 files and zero alterations.
- Confirmed every linked library resolves with the declared dependencies installed.
- Verified normal sync holds the fresh release, while `BYPASS_MIN_RELEASE_AGE=1` recognizes `0.6.2` and produces no package diff.
- Merged the PR head with current upstream in a disposable checkout and rebuilt successfully.
- Exact repository-builder jobs passed on [edge, rc, and stable](https://github.com/DanWahlin/omarchy-pkgs/actions/runs/34815698486).
- All 12 native Rust tests passed, including marker detection. The browser-only suite, which does not exercise this Rust change, passed 214/236 under the local system Chromium; the remaining timing-sensitive game cases were treated as non-blocking.
- Ran the `sync-upstream`, `omarchy-pkgs`, and `omarchy-release` self-tests successfully.

Upstream's desktop file is renamed from `Agent Arcade.desktop` to `agent-arcade.desktop`; its `Exec=agent-arcade` entry remains unchanged.

I'm the upstream author and will maintain this package.


Chessing234 (approved):
deb repackage with sha256 on license + asset. upstream sync config looks right.

PR #340, opened 2026-09-07
fix(flea): detect O_NOFOLLOW after the 0.1.4 refactor

thisisgm:
The b4b7ee4 check greps `copyfile.rs` for a literal `.custom_flags(O_NOFOLLOW)` that Flea 0.1.4 moved behind `regfile::open_if_regular`, so `upstream.sh` has refused every release since and flea is pinned at 0.1.3 -> both sites now accept either spelling, verified present on v0.1.3, v0.1.4 and v0.1.5.


thisisgm:
Superseded by #469, which carries the same check against the 0.3.0 line.

PR #341, opened 2026-09-07
Fix ARM Snapper dependency and first-run keyboard service packaging

scottjones:
## Summary

- Require Snapper on aarch64 as well as x86_64 in both desktop recipes; only the Limine stack remains x86-specific. Fresh btrfs Apple Silicon installs need Snapper for root configuration.
- Install `omarchy-brightness-keyboard-auto.service` into systemd's user-unit directory from both settings recipes when the source contains it. Current Mac first-run setup requires this unit; its omission prevents the completion marker and repeats the notification stack.
- Keep the existing stable source pin buildable: it does not contain this newer unit, so installation is conditional. Source pins and release versions are unchanged.

## Verification

- Shell syntax and `git diff --check` pass.
- The companion desktop regression test executes the real package functions for all eight stable/development × aarch64/x86_64 × current-source/actual-stable-pin combinations. All pass; current-source cases verify every unit requested by the real first-run leaf.
- Full ISO/rootfs rebuild with local desktop/settings packages `4.0.2-5` and Snapper `0.13.1-3`, followed by a fresh encrypted M3 (j613) install: Snapper configuration and factory snapshot present, first-run marker complete, required user services enabled. Tester subsequently confirmed reboot and no repeated alerts.
- This is source for review, not a production release/publish operation; the local test build supplied its own package release override.

## Related work

- Desktop companion: https://github.com/omarchy-mac/omarchy-mac/pull/372 (regression test and explicit Snapper setup diagnostics).
- Installer companion: https://github.com/omarchy-mac/omarchy-mac-iso/pull/19.
- Related but separate desktop repair/build-wrapper approach: https://github.com/omarchy-mac/omarchy-mac/pull/367. This PR fixes the underlying upstream recipes; it does not change that PR or its migration.

### DIFF
--- a/pkgbuilds/omarchy-dev/PKGBUILD
+++ b/pkgbuilds/omarchy-dev/PKGBUILD
@@ -44,6 +44,7 @@ depends=(
 
   # System font (UI assumes it)
   'ttf-jetbrains-mono-nerd-basic'
+  'snapper' # Root snapshots also apply to btrfs installations on Apple Silicon.
 )
 
 # Bootloader stack. Lives here rather than on omarchy-settings because
@@ -57,11 +58,10 @@ depends_x86_64=(
   'limine'
   'limine-mkinitcpio-hook'
   'limine-snapper-sync'
-  'snapper'
 )
 
 # Apple Silicon boots through m1n1 + GRUB from the Asahi packages and keeps
-# Arch Linux ARM's kernel and boot layout, so the Limine/Snapper stack does not
+# Arch Linux ARM's kernel and boot layout, so the Limine stack does not
 # apply. Wi-Fi on the Broadcom parts only scans reliably through the iwd
 # backend, which install/hardware/network.sh selects on Apple Silicon.
 depends_aarch64=(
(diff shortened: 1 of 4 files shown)


malik-na (changes requested):
The Snapper dependency and conditional keyboard-unit changes look correct. Please resolve these two surrounding packaging issues here or in an explicit prerequisite before using this package set for the Mac rollout:

1. Preserve the historical ARM mkinitcpio drop-ins with native defaults and pacman backup protection. Upgrading from 4.0.2-2 currently removes package-owned boot policy, losing Plymouth/console configuration and changing keyboard ordering on the next initramfs rebuild. This exclusion predates this PR; an unbootable machine was not demonstrated.
2. Guard the integrated development recipe's plocate drop-in installation, as stable already does. Its unconditional install fails against the tested Mac/older source trees where that file is absent. This is independent of the keyboard-unit addition.

Local fixes are implemented for fresh installs and upgrades, including customized-file preservation. Native ARM builds, real pacman lifecycle checks and disposable ARM VM tests passed; the boot-policy checks were rerun successfully. Physical Apple boot/recovery acceptance remains outstanding. No source changes have been pushed.

The PR head and current master still match the tested revisions. Full evidence and fix details: https://github.com/omacom/omarchy-pkgs/pull/341#issuecomment-5618609454

PR #349, opened 2026-09-08
Add tb-ar-shim-dkms: Alpine Ridge suspend fix for the MacBookPro14,1

orospakr:
Adds `tb-ar-shim-dkms`, a temporary (`[FOR-UPSTREAM]`) DKMS package fixing suspend/resume of the Alpine Ridge Thunderbolt 3 controller on the Apple MacBookPro14,1, per the investigation in omacom/omarchy#10593.

**Shape:** the same as `qmk-hid` / `tzupdate` — a `"source": "local"` PKGBUILD that fetches a tag tarball from the module's own public repo, https://github.com/orospakr/linux-apple-alpine-ridge-suspend, with a `git_tags` upstream provider in `package.json` so `sync-upstream` opens the version bumps. What omarchy-pkgs carries is the PKGBUILD and the metadata (~35 lines); the module source, its README, the in-tree `drivers/pci/quirks.c` patch headed to linux-pci, and the CI that builds the module against Arch `linux`/`linux-lts` headers and re-checks the upstream patch against pci-next weekly all live in that repo. (An earlier revision of this PR vendored the C source here; this is the slimmed-down replacement.)

**What it fixes:** two kernel bugs on this machine, both of which leave Thunderbolt and the USB-C ports dead after every suspend:

1. **deep/S3:** the PCI core saves the controller's bridge config space *after* the thunderbolt driver has already put the chip to sleep, so it saves all-ones and restores garbage on wake — 75–107 s resumes, dead NHI, hot-removed xHCI.
2. **s2idle:** the core skips config restore entirely (`skip_bus_pm` + `pm_suspend_no_platform()`), but this controller genuinely loses its config space during sleep — dead NHI (`-ESHUTDOWN`) until module reload.

The module grafts a `dev_pm_domain` onto the controller's upstream bridge: saves the subtree's config space in `suspend_late` (while the chip is still readable) and restores it parents-first in `resume_noirq`. It autoloads by PCI modalias and verifies the on-board controller via Apple's ACPI signature (same trick as the mainline Cactus Ridge quirk), so it exits `-ENODEV` everywhere else.

**Validation:** 28/28 suspend cycles on a MacBookPro14,1 across both sleep modes, 1.3–1.4 s resumes; the previous revision of this exact module was installed via pacman, dkms-built, and verified through a cold boot (modalias autoload + suspend). `makepkg` of this revision verified locally against the v1.0.0 tag tarball; the module source is byte-identical apart from one comment line pointing at the new repo.

**Sunset:** documented in the upstream repo README — once the quirks land in mainline and every bootable kernel has them, the package is removed with `pacman -Rns`. The companion install-script/migration PR omacom/omarchy#10758 (gating on the exact DMI model and defaulting the machine to s2idle) depends on this package being published.

https://claude.ai/code/session_014SjBV8RAyo6et6LMmurU9W

### DIFF
--- a/pkgbuilds/tb-ar-shim-dkms/.omarchy/package.json
+++ b/pkgbuilds/tb-ar-shim-dkms/.omarchy/package.json
@@ -1,0 +1,10 @@
+{
+  "source": "local",
+  "upstream": {
+    "git_tags": "https://github.com/orospakr/linux-apple-alpine-ridge-suspend.git",
+    "tag_pattern": "v{pkgver}",
+    "sources": {
+      "any": ["https://github.com/orospakr/linux-apple-alpine-ridge-suspend/archive/refs/tags/{tag}.tar.gz"]
+    }
+  }
+}
--- a/pkgbuilds/tb-ar-shim-dkms/PKGBUILD
+++ b/pkgbuilds/tb-ar-shim-dkms/PKGBUILD
@@ -1,0 +1,20 @@
+# Maintainer: Andrew Clunis <andrew@orospakr.ca>
+_pkgbase=tb-ar-shim
+_repo=linux-apple-alpine-ridge-suspend
+pkgname=tb-ar-shim-dkms
+pkgver=1.0.0
+pkgrel=1
+pkgdesc="Suspend/resume fix for the Alpine Ridge Thunderbolt controller on Apple MacBookPro14,x (DKMS)"
+arch=('x86_64')
+url="https://github.com/orospakr/$_repo"
+license=('GPL-2.0-only')
+depends=('dkms')
+source=("$pkgname-$pkgver.tar.gz::$url/archive/refs/tags/v$pkgver.tar.gz")
+sha256sums=('30ceacceabe854904ef4d555e225e9901c17fd7b524f075c4a17de8e7b2a8b9c')
+
+package() {
+  cd "$_repo-$pkgver"
+  install -Dm644 -t "$pkgdir/usr/src/$_pkgbase-$pkgver" \
+    module/tb_ar_shim.c module/Makefile module/dkms.conf
+  install -Dm644 -t "$pkgdir/usr/share/doc/$pkgname" README.md
+}


angellindo:
**Data point from a MacBookPro14,3** (dual Alpine Ridge JHL6540, kernel `7.2.5-3-omarchy`, s2idle). I have not run this shim. This is about scope and an alternative, not a test result for this package.

- **Same hardware path as the 14,1.** Both upstream bridges here are `8086:1578` (`04:00.0` behind `PEG1`, `7a:00.0` behind `PEG2`). `SXFP` exists as `\_SB.PCI0.PEG1.UPSB.DSB0.NHI0.SXFP` and the `PEG2` equivalent, in SSDT7 rather than the DSDT. So the shim's device match would likely bind here. The detector in [omarchy#10758](https://github.com/omacom/omarchy/pull/10758) limits it to 14,1.
- **Same failure.** Stock s2idle on the 14,3 shows your signature. At suspend the switch ports are inaccessible before D3hot. On resume the NHI and xHCI are `not ready` and `tb_cfg` returns `-108`.
- **Alternative for s2idle.** Instead of restoring the config that `tb_lc_set_sleep()` clears, don't set the LC sleep bit on the host router when the target is s2idle. That is patch `0664` in [omarchy-pkgs#596](https://github.com/omacom/omarchy-pkgs/pull/596). Tested as a kprobe on the stock module, it kept both domains over six sleeps in one boot. It does not cover S3, where this patch leaves the old path.
- **Question.** With the LC sleep skipped, noirq resume here is still about 17.5 s: 11.5 s and 12.6 s in `pci_pm_resume_noirq` on the two upstream bridges, and about 5.5 s per NHI. Your 1.3–1.4 s resumes are much faster. Were they measured with or without the ACPICA fix from [omarchy-pkgs#544](https://github.com/omacom/omarchy-pkgs/pull/544)? And does the pm_domain on the upstream bridge change when that firmware code runs?
- **Test method.** Here only the first sleep after a boot tells anything about Thunderbolt. Once one sleep fails, every later one starts with `-108`.

PR #357, opened 2026-09-08
Fix IPU7 camera rebuilds across jsoncpp ABI changes

### LINKED ISSUE #10837: RC jsoncpp 1.9.8 soname bump breaks intel-ipu7-camera 1.0.5
### System details

- Dell XPS 14 DA14260, BIOS 1.8.2
- Intel Core Ultra X7 358H (Panther Lake)
- Intel Arc B390 (`xe`)
- Omarchy stable: `4.0.2-1`
- Omarchy RC tested: `4.0.3rc3-1`
- Kernel: `7.1.8-arch1-2-ptl`
- Camera: IPU7 `8086:b05d`, OV08X40 / `OVTI08F4`
- `intel-ipu7-camera 1.0.5-1`
- `v4l2-relayd 0.2.0-2.2`
- Stable `jsoncpp 1.9.6-3` provides `libjsoncpp.so.26`
- RC `jsoncpp 1.9.8-1` provides `libjsoncpp.so.27`

### What's wrong?

Switching from stable to the RC repository and updating to Omarchy `4.0.3rc3-1` upgrades `jsoncpp` from `1.9.6-3` to `1.9.8-1`. That removes `libjsoncpp.so.26` and installs `.27`, but the RC repository still serves `intel-ipu7-camera 1.0.5-1`, whose packaged IPU7 HAL plugin is linked to `.26`:

```text
$ readelf -d <intel-ipu7-camera-1.0.5-1>/usr/lib/libcamhal/plugins/ipu75xa.so | grep jsoncpp
Shared library: [libjsoncpp.so.26]
```

The result is that the built-in IPU7 camera stops working after the RC update. The kernel side remains healthy and `/dev/video50` exists, but `icamerasrc` cannot load the HAL and the relay produces no usable stream.

### Timeline and correction

The initial version of this report incorrectly said the RC update itself rewrote the HAL binaries to require `.27`. Further journal inspection showed that was not true. The actual timeline was:

1. At 09:18, the RC transaction upgraded `jsoncpp 1.9.6-3 -> 1.9.8-1`, while `intel-ipu7-camera` remained `1.0.5-1` linked to `.26`. This is when the initial RC incompatibility was introduced.
2. At 15:45, during local troubleshooting, a manually rebuilt HAL was explicitly copied over the package-owned files. That local rebuild linked against the then-current `.27` ABI.
3. At 18:43, downgrading to stable restored `jsoncpp 1.9.6-3` (`.26`), but naturally did not restore those manually replaced package files. The local `.27` rebuild therefore remained broken after downgrade.
4. Reinstalling `intel-ipu7-camera` restored the packaged `.26` binaries and fixed the camera on stable.

The manual rebuild explains why `pacman -Qkk` later reported altered files; it was not performed by Omarchy. Sorry for the incorrect attribution in the original report.

### Initial RC failure

With the stock camera package and RC `jsoncpp 1.9.8-1`, the dependency relationship is:

```text
intel-ipu7-camera 1.0.5-1 ipu75xa.so -> libjsoncpp.so.26
RC jsoncpp 1.9.8-1                    -> libjsoncpp.so.27 only
```

The camera HAL therefore cannot load unless `intel-ipu7-camera` is rebuilt for the RC ABI or a compatible `.26` library remains available.

### Steps to reproduce

1. Start on stable Omarchy `4.0.2-1` with working `intel-ipu7-camera 1.0.5-1` and `jsoncpp 1.9.6-3`.
2. Switch to the RC repository and update to `4.0.3rc3-1`.
3. Confirm `jsoncpp` is now `1.9.8-1` and only `libjsoncpp.so.27` exists.
4. Check the packaged camera plugin with `ldd` or `readelf`; it still requires `libjsoncpp.so.26`.
5. Try the camera in a browser or with the V4L2 relay; the HAL cannot load and no usable frames are produced.

### Stable recovery verification

After returning to stable, reinstalling the camera package restores the matching ABI:

```bash
pkexec pacman -S intel-ipu7-camera
pkexec systemctl restart v4l2-relayd@ipu7.service
```

```text
$ ldd /usr/lib/libcamhal/plugins/ipu75xa.so | grep jsoncpp
libjsoncpp.so.26 => /usr/lib/libjsoncpp.so.26

$ pacman -Qkk intel-ipu7-camera
intel-ipu7-camera: 834 total files, 0 altered files
```

A V4L2 probe then captured 10 complete 1920x1080 NV12 frames (`31,104,000` bytes), and applications detected the camera again.

### Expected fix

Publish an RC build/revision of `intel-ipu7-camera` linked against `libjsoncpp.so.27`, or otherwise keep the required `.26` ABI available in the RC package set. The camera package may also benefit from a versioned soname dependency so repository validation catches the mismatch before publication.

iuliansafta:
## Problem

The RC channel upgraded `jsoncpp` from 1.9.6 (`libjsoncpp.so.26`) to 1.9.8 (`libjsoncpp.so.27`), while `intel-ipu7-camera 1.0.5-1` remained linked to `.26`. `ipu75xa.so` could no longer load, leaving the IPU7 V4L2 relay present but unable to produce frames.

## Fix

- Mark `intel-ipu7-camera` as a fast-ring package so it is built against each channel's own dependency set rather than carrying an artifact across incompatible channel ABIs.
- Track `jsoncpp` through `rebuild_on`, allowing the existing rebuild automation to bump the package when that dependency moves.
- Depend on the `libjsoncpp.so` soname provider rather than only the unversioned package name, exposing ABI compatibility to package metadata.
- Bump `pkgrel` to 2 so corrected artifacts are published now.

Fixes omacom/omarchy#10837.

## Verification

- `./bin/sync-rebuilds intel-ipu7-camera` generated the pkgrel bump and x86_64 dependency record.
- `./bin/sync-rebuilds --self-test`
- `./bin/omarchy-pkgs self-test`
- `bash -n pkgbuilds/intel-ipu7-camera/PKGBUILD`
- `makepkg --printsrcinfo` reports `depends = libjsoncpp.so`.
- `git diff --check`

A full IPU7 package build is intentionally left to the repository's native per-channel builders because the package includes kernel DKMS modules and proprietary camera components.

### DIFF
--- a/pkgbuilds/intel-ipu7-camera/.omarchy/package.json
+++ b/pkgbuilds/intel-ipu7-camera/.omarchy/package.json
@@ -1,3 +1,12 @@
 {
-  "source": "local"
+  "source": "local",
+  "release_ring": "fast",
+  "rebuild_on": [
+    "jsoncpp"
+  ],
+  "rebuilt_against": {
+    "x86_64": {
+      "jsoncpp": "1.9.6-3"
+    }
+  }
 }
--- a/pkgbuilds/intel-ipu7-camera/PKGBUILD
+++ b/pkgbuilds/intel-ipu7-camera/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=intel-ipu7-camera
 pkgver=1.0.5
-pkgrel=1
+pkgrel=2
 pkgdesc="Intel IPU7 MIPI camera stack for Hurrican/Performance (OV08X40 + hardware ISP)"
 arch=('x86_64')
 url="https://github.com/TsaiGaggery/hurrican_omarchy_enabling"
@@ -16,7 +16,7 @@ depends=(
   'gst-plugins-base'
   'gst-plugins-good'
   'gst-plugins-bad'
-  'jsoncpp'
+  'libjsoncpp.so'
   'libdrm'
 )
 makedepends=(


ryanrhughes:
Closing: master already carries rebuild_on jsoncpp for intel-ipu7-camera and the package has been rebuilt against jsoncpp 1.9.8, so the ABI break this fixed is handled. Thanks for the report and the fix.

Issue #366, opened 2026-09-09
intel-ipu7-camera: Panther Lake-only build breaks Lunar Lake — no camera at all, and the image is upside down


pkolbas:
`intel-ipu7-camera` carries two Panther Lake assumptions that break on Lunar
Lake. Both are one-line fixes, and I have verified each on hardware.

**Machine:** Dell Pro 14 Premium PA14250 — `Core Ultra 200V Series Processors
IPU [8086:645d]` (Lunar Lake), bridge `INTC10DE`, sensors `OVTI08F4` (OV08X40,
RGB) + `HIMX1092` (IR). Omarchy 4.0.3-1, kernel 7.2.3-arch1-3. The package was
installed by the installer at first boot, so this is what the machine ships with.

---

## 1. The HAL plugin is built for `ipu75xa` only, so there is no camera at all

`pkgbuilds/intel-ipu7-camera/PKGBUILD:117`

```bash
-DIPU_VERSIONS="ipu75xa" \
```

`libcamhal` picks its plugin by platform, and on Lunar Lake it asks for the one
that was never built:

```
CamHAL[ERR] HalAdaptor: load_camera_hal_library, failed to open library:
  /usr/lib/libcamhal/plugins/ipu7x.so: No such file or directory
CamHAL[ERR] HalAdaptor: get_number_of_cameras, function call is nullptr
```

`/usr/lib/libcamhal/plugins/` contains only `ipu75xa.so`, and `/etc/camera/`
only `ipu75xa/` with PTL tunings (`OV08X40_KAFE799_PTL.aiqb`). The result is a
camera that cannot work at all — `icamerasrc` loads and then reports
`parseSensors: No sensors available`.

Worth noting: the package **already ships** the whole matching algorithm set
from `ipu7-camera-bins` — `libia_aic-ipu7x.so`, `libia_cca-ipu7x.so`,
`libia_dvs-ipu7x.so` and the rest are all installed. Only the HAL plugin and
its config directory are missing.

**Fix:** `-DIPU_VERSIONS="ipu7x;ipu75xa"`. The upstream tree builds both from
one configure run, and the AUR `intel-ipu7-camera-hal-git` already does exactly
this. The `ipu7x` config set that comes with it carries the right tuning for
this board: `OV08X40_BBG802N3_LNL.aiqb` and
`gcss/OV08X40_BBG802N3_LNL.IPU7X.bin`.

[...]

PR #367, opened 2026-09-09
intel-ipu7-camera: build the Lunar Lake HAL plugin and stop rotating its frames

pkolbas:
Fixes items 1 and 2 of #366, now on top of 1.0.6-2 (#418).

**Rebased 2026-09-16.** #418 turned out to do the kernel half for Lunar Lake as
well: with this package's own `intel-cvs` DKMS module and the in-tree
`ipu-bridge`, the OV08X40 on this board joins the media graph behind the
`Intel CVS` entity exactly as it does on Panther Lake. What is still missing is
the user-space half, and that is this PR.

### 1. Build both HAL platforms, with the CVS pad formats for `ipu7x`

`libcamhal` picks its plugin by platform, and on Lunar Lake it asks for one that
was never built:

```
CamHAL[ERR] HalAdaptor: load_camera_hal_library, failed to open library:
  /usr/lib/libcamhal/plugins/ipu7x.so: No such file or directory
CamHAL[WAR] CameraParserInvoker: parseSensors: No sensors available
```

The proprietary half is already shipped — `libia_aic-ipu7x.so` and the rest of
that set come from `ipu7-camera-bins` today. Upstream builds both platforms from
one configure run, and the existing `make DESTDIR="${pkgdir}" install` lays down
`/etc/camera/ipu7x/` next to `ipu75xa` on its own, tuning included — so the
build change is one entry in `IPU_VERSIONS`, with the explicit `ipu75xa` config
installs left untouched.

New in this revision: `0008-camhal-ipu7x-ov08x40-Intel-CVS-formats.patch`, the
`ipu7x` twin of #418's `0006`. Same hunk, same reason: behind the 7.2 bridge the
`Intel CVS` pads need a format. `0005` alone is not enough for `ipu7x` either —
without the routing patch the plugin resolves the bridge's name as the I2C bus
(`I2CBus:S`, `failed to setup Link ov08x40 S ==> Intel IPU7 CSI2 0`) and relayd
delivers one black frame; with `0005` + `0008` it streams.

### 2. Pick the relay pipeline per board

`v4l2-relayd-ipu7.conf` rotates every frame 180°. That is right where the sensor
is mounted inverted, and wrong here — this sensor says so itself:

```
camera_orientation      0x009a0922 (menu) : value=0 (Front)   read-only
camera_sensor_rotation  0x009a0923 (int)  : value=0           read-only
```

so the picture arrives upside down in every application that opens
`/dev/video50`.

`camera-init.service` now runs a small script that writes `VIDEOSRC` to
`/run/v4l2-relayd-ipu7-videosrc.env` based on the bridge ACPI id, and the relay
drop-in reads it. **Only `INTC10DE` takes the new path.** `INTC10E1`,
`INTC10CF`, `INTC10E0` and anything unrecognised keep the packaged pipeline byte
for byte, and because the drop-in marks the file optional, a boot where
`camera-init` did not run falls back to it too. I have no Panther Lake hardware,
so the change is shaped to be a no-op there by construction rather than by my
testing.

The ordering works because the relay drop-in already carries
`After=camera-init.service`, and because systemd applies every `EnvironmentFile`
after every `Environment=`, with drop-in files read after the unit's own — so
this one wins over `/etc/v4l2-relayd.d/ipu7.conf` without that file being
touched.

### Verified

Dell Pro 14 Premium PA14250 — Lunar Lake, `INTC10DE`, Synaptics SVP7500
`06cb:0701`, `OV08X40` + `HM1092`; `intel-ipu7-camera 1.0.6-2`,
`linux-omarchy 7.2.5-3`:

* `intel-cvs/1.0.6` DKMS from this package, in-tree `ipu-bridge`: `Intel CVS`
  entity between `ov08x40 17-0036` and `Intel IPU7 CSI2 0`; `intel_cvs
  i2c-INTC10DE:00: Quirks: 0x3a (VID:0x06cb PID:0x0701)`
* `ipu7x` plugin from the pinned commit with `0005` + `0008`:
  `resolveCsiPortAndI2CBus: I2CBus:17-0036 <=> CSI Port:0`,
  `(0): ov08x40-uf - ov08x40 as sensor.`
* `v4l2-relayd@ipu7`: 30 fps on `/dev/video50`, live picture, upright in a
  browser
* `makepkg` of this branch builds; the `ipu7x` plugin and `/etc/camera/ipu7x/`
  it produces are what runs on this machine

### Still not sufficient on Lunar Lake — outside this PR, written up in #366

1. **Sensor probe races the bridge.** The sensor's I2C bus is handed to the host
   only in the bridge's runtime resume; `camera-init`'s `modprobe intel_cvs &&
   sleep 2 && modprobe ov08x40` lands while the bus is still owned by the bridge
   firmware, and `ov08x40` fails with `-110` on the chip-id read. Fix shape:
   probe the sensor with the bridge held awake (drop-in tested here, boot and
   resume). #368 rewrites the same service for a related resume problem.
2. **Privacy LED never lights.** The in-tree `cvs` driver's quirk table gives
   `06cb:0701` `ICVS_HOST_PRIV_CTRL`, so the bridge hands the LED to a host that
   has none. Dropping that flag (keeping `HOST_SENSOR_PWR_CTRL`) brings it back;
   belongs upstream, could be carried like `0007` meanwhile.
3. **IR sensor.** `ipu-bridge` learns `HIMX1092` in 7.3 (`4fdb0342f05e`), one
   hunk for `linux-omarchy`. With it `hm1092` links straight to CSI2 port 1 and
   streams 30 fps.
4. **psys on s2idle.** `psys_suspend` returns `-EBUSY` after the camera was
   used, so suspend fails; platform-independent, comes from Intel's tree, the
   community pack carries a patch that applies to 1.0.6 cleanly.

I carry all four locally and can send each as its own PR.


pkolbas:
Rebased on top of 1.0.6-2 (#418), which changed the picture: this package's own `intel-cvs` DKMS module plus the in-tree `ipu-bridge` bring the Lunar Lake sensor into the graph behind `Intel CVS` too, so the kernel half is done and this PR is now the whole user-space half.

Two changes from v1: `0008`, the `ipu7x` twin of `0006` (CVS pad formats, same hunk, same reason), and the description, which now lists what is still missing on this board and why. Verified against 1.0.6-2 on `linux-omarchy 7.2.5-3`: graph, HAL, 30 fps, upright image, `makepkg` builds.

@spencerbull — #418's description lists #366 as "not addressed, different platform". This is the addressed half; #366 has the rest, with evidence and a concrete question.

PR #380, opened 2026-09-10
omarchy-settings: keep the Limine and mkinitcpio drop-ins on aarch64

jdvmi00:
The aarch64 omarchy-settings packages remove etc/mkinitcpio.conf.d and
etc/limine-entry-tool.d wholesale, on the assumption that aarch64 means
Apple Silicon without Limine. The aarch64 installs that exist today, the
DGX Spark and the Snapdragon X laptops (#8672 / omarchy-iso#156), boot with
Limine and an encrypted root exactly like x86_64.

What happens on such an install: installing the published package drops
omarchy_hooks.conf, so HOOKS falls back to /etc/mkinitcpio.conf. Nothing
breaks until the next initramfs build. The first kernel update then produces
an image whose hook run order is `resume` only, with no encrypt hook, no
dm-crypt and no plymouth, and boot hangs on "A start job is running for
/dev/mapper/root" with no prompt. Reproduced on a DGX Spark on edge after
omarchy-settings-dev 4.0.0.r2015.g4930677-1 and linux-aarch64 7.2.3-2 →
7.2.4-1; verified with lsinitcpio; restoring omarchy_hooks.conf and running
limine-update fixed it.

Change: ship both drop-in directories, the Limine template and the snapper
notifier on aarch64 too, and remove only thunderbolt_module.conf, whose
module ARM kernels lack. Memory-stack removals are unchanged. Applied to
omarchy-settings and omarchy-settings-dev alike; omarchy-settings pkgrel
bumped.

Testing: built omarchy-settings-dev for aarch64 with bin/build in the repo
container on the Spark; the package lists etc/mkinitcpio.conf.d/omarchy_hooks.conf
and etc/limine-entry-tool.d/*.conf, and none of the zram/oomd/zswap files.
An Apple Silicon install without Limine would now get omarchy_hooks.conf,
which names btrfs-overlayfs and plymouth; that path (#9835) is unmerged and
can gate the hooks it does not want when it lands.

### DIFF
--- a/pkgbuilds/omarchy-settings-dev/PKGBUILD
+++ b/pkgbuilds/omarchy-settings-dev/PKGBUILD
@@ -6,10 +6,11 @@ _pkgver_base=4.0.0
 _pkgver_base_tag=v3.8.2
 pkgdesc='Omarchy user defaults, /etc/skel content, fonts, plymouth theme, and support helpers (quattro branch tip)'
 # Arch-specific because the shipped /etc tree is not the same on every
-# architecture: the Limine, mkinitcpio, zram and oomd drop-ins belong to the
-# x86_64 boot and memory stack and are left out of the aarch64 package (see
-# package()). makepkg only honours the arch-suffixed arrays below on
-# arch-specific packages.
+# architecture: the zram and oomd drop-ins belong to the x86_64 memory stack
+# and are left out of the aarch64 package (see package()). The Limine and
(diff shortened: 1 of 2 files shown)


birkskyum:
@jdvmi00 , there's a pr here against your branch that updates this pr.
- https://github.com/jdvmi00/omarchy-pkgs/pull/1

PR #382, opened 2026-09-10
fix(camera): complete the native CVS package transition

rfrost-xyz:
Relates to omacom/omarchy#10948.

Linux 7.2 [routed IPU camera discovery through CVS](https://github.com/torvalds/linux/commit/c6b1b34b509032c7e7cef9efc63cab55c2ad309e), but the packaged kernel leaves the native CVS driver disabled. The legacy DKMS driver does not register the required V4L2 endpoint, leaving IPU7 waiting for `INTC10E1-0/port@1/endpoint@0` and OV08X40 unavailable.

This proposes a coordinated native CVS transition in the kernel and camera packages:

- Enable native CVS in `linux-ptl 7.2.3.arch1-2`. Add its missing ancillary autoselection default while preserving the upstream helper-menu placement and dependency bounds. Fail package preparation if `olddefconfig` drops the driver.
- Carry the existing firmware-owned sensor-power policy into the native driver specifically for Synaptics `06cb:0701` with Panther Lake `INTC10E1`. Preserve privacy ownership and behaviour on other devices.
- Release camera stack `1.0.6-1` with native-kernel exclusion for legacy CVS DKMS. The versioned source-directory change lets the existing package removal hook retire the old registration. Legacy kernels retain the legacy driver.
- Backport Intel's [native-CVS HAL topology support](https://github.com/intel/ipu7-camera-hal/commit/f167239b3ecf242ae081767892cb0a4c54bad496), propagate negotiated formats to the bridge, report format failures and add the required bundled UAPI definitions. Preserve existing sensor configuration and direct-sensor support.

Validation at `2e00e40`:

- Full kernel, headers and camera package builds pass in isolated edge containers.
- Real package transactions verify native CVS ownership after a combined upgrade, then restore the exact baseline packages and legacy DKMS registration on reversal. Camera-only upgrade/reversal also preserves legacy-kernel support.
- Full kernel configuration, 11 Kconfig dependency cases, HAL methods using bundled headers and nine power/privacy-policy cases pass. Regression checks fail against the corresponding original sources. These tests were run manually; the current CI workflow does not invoke them.

**Ready for source review; hardware validation is required before merge.** This is a proposed fix, not a confirmed camera repair. The affected daily-use laptop has not been upgraded to these candidate packages. A maintainer or affected-device tester is needed to validate boot, capture, privacy and suspend/resume. Container transactions exercise package/DKMS behaviour and initramfs generation, but not Limine/UEFI boot, actual camera frames, browser capture, privacy indication or suspend/resume. The logs include expected legacy-DKMS exclusion and container-only systemd, vconsole and unrelated header-only kernel warnings. The proposed Kconfig default and native power policy have not been accepted upstream or verified on the affected hardware.


rfrost-xyz:
This is an unverified implementation candidate, not a confirmed camera repair. Before testing it on the affected XPS, could maintainers confirm:

- Is completing the native CVS transition the intended direction for the packaged IPU7 stack, or is another fix already underway?
- Does the coordinated kernel, legacy-DKMS handover and HAL approach here fit that direction?
- Should the Panther Lake firmware-owned sensor-power policy be carried into the native driver here, or handled differently and submitted upstream first?

The focused tests pass and the native module compiles, but complete package builds, an upgrade/reversal rehearsal and hardware acceptance remain outstanding. I am holding installation pending feedback on the approach and keeping this PR draft. If an official fix supersedes it, I will align with that rather than maintain a separate local stack.

jzetterman:
Data point from a Dell XPS 16 DA16260 (SKU 0DBA, audio subsystem `1028:0dba`), a different SKU from the XPS 14 in omacom/omarchy#10948. The in-tree CVS driver, as it stands in every v7.2.x tag through v7.2.6, kills the internal speakers on this machine the same way #418 found on the DA14260. This PR's linux-ptl patches (0030, 0031) do not carry the fix that #418 applied to its DKMS copy, `0007-media-i2c-cvs-get-the-wake-irq-without-claiming-the-gpio.patch` (Junjie Cao, Cc: stable, not yet in stable). If linux-ptl builds the driver in-tree, it needs that patch too, or the amps fail on at least these two SKUs.

Caveat: I did not run this PR's packages. I built the unmodified `drivers/media/i2c/cvs` from v7.2 as a local DKMS module on `linux-ptl 7.2.3.arch1-1` with the vision-drivers module removed. That is the same driver source this PR enables in the kernel. I cannot speak for other SKUs.

**What fails**

When `intel_cvs` probes before the amps, all four CS35L57 SoundWire amps fail and no sound card registers at all, not even HDMI:

```
[   26.716034] kernel: intel_cvs i2c-INTC10E1:00: Quirks: 0x7a (VID:0x06cb PID:0x0701)
[   27.428666] kernel: cs35l56 sdw:0:2:01fa:3557:01:2: error -EBUSY: Failed to get spk-id-gpios
[   27.428902] kernel: cs35l56 sdw:0:2:01fa:3557:01:2: probe with driver cs35l56 failed with error -16
```

The same two lines repeat for `sdw:0:2:01fa:3557:01:3`, `sdw:0:3:01fa:3557:01:0` and `sdw:0:3:01fa:3557:01:1`. `/proc/asound/cards` stays empty. Kernel `7.2.3-arch1-1-ptl`, linux-firmware 20260810-2, sof-firmware 2025.12.2-1. When the amps do probe, they load `cs35l57-b2-dsp1-misc-10280dba-spkid0-*` tuning.

These are SoundWire amps (`sdw:0:2:01fa:3557`), not the SPI sidecar amps in the XPS 13 threads (omacom/omarchy#10987, #9687, #10543). Those fail on tuning files and the sidecar quirk. This fails on a GPIO request and never gets that far.

**Why**

`/sys/kernel/debug/gpio` with `intel_cvs` loaded:

```
gpiochip1: 51 GPIOs, parent: platform/INTC10BC:01, INTC10BC:01:
 gpio-20  (                    |wake                ) in  lo IRQ ACTIVE LOW
 gpio-36  (                    |rst                 ) out hi
```

`INTC10BC:01` is `\_SB.GPI1`. Pin 20 is the pin #418 identified as the amps' shared `spk-id-gpios` on the DA14260. Same pin here. The driver takes it in `core.c` with `devm_gpiod_get(dev, "wake", GPIOD_IN)` and holds it while bound. `cs35l56_get_speaker_id()` takes `spk-id` with `gpiod_get_array_optional()`, reads it, and releases it with `gpiod_put_array()`. `-EBUSY` is not `-EPROBE_DEFER`, so the amp probe fails once and never retries. The old vision-drivers `intel_cvs` requests only `req`, `resp` and `rst`, never `wake`, so 7.1 with the old driver was fine here in either probe order.

**It is a race, so one clean boot does not clear it**

- Boot on 2026-09-13 (EDT): amps probed at 15.25 s, `intel_cvs` at 15.49 s. Speakers worked.
- Boot on 2026-09-15: `intel_cvs` at 26.72 s, amps at 27.43 s. No sound card.

A validator who boots once and hears audio has not ruled this out. Check `dmesg | grep spk-id` on every boot.

**Workaround I ran before finding #418**

Blacklist `intel_cvs` in modprobe.d so udev does not autoload it, and add an `ExecStartPre` to `camera-init.service` that waits for `sof-soundwire` in `/proc/asound/cards` before the unit's own `modprobe intel_cvs`. Because the amps release the line after reading it, amps-first works for both drivers. Camera and speakers both come up. The wake-IRQ patch is the right fix. This only shows the mechanism and gives a validator a way to get audio back without rebuilding.

PR #390, opened 2026-09-11
Add Ghost: runtime with optional Omarchy HUD

ferdousbhai:
Adds Ghost 0.4.0 as two packages: `ghost-runtime` contains the daemon, CLI, opt-in Chromium relay, and desktop automation; `ghost` adds the Omarchy HUD and depends on the exact matching runtime version. Apps can install only `ghost-runtime` and provide their own UI in a Hyprland graphical session, without a Quickshell dependency.

Both packages build from the same sanitized source and portable Bun runtime archives in https://github.com/ferdousbhai/ghost/releases/tag/v0.4.0. The runtime archive is 1.28 MB, down 44% from 0.3.0. The pinned Pi patch removes unused executable file-extension loading; native inline extensions, image processing, and bundled license files remain intact.

No network build step. `.omarchy/package.json` follows the GitHub release with digests and versioned source URLs. Local `makepkg --cleanbuild --noconfirm` completed against the published inputs and produced `ghost-runtime-0.4.0-1-any.pkg.tar.zst` and `ghost-0.4.0-1-any.pkg.tar.zst`. The full Ghost gate and package ownership/dependency checks passed.

The matching Install → AI → Ghost and Remove → AI → Ghost UI actions are in https://github.com/omacom/omarchy/pull/11250. UI removal preserves the installed runtime and owner data.


ferdousbhai:
Updated to ghost 0.3.0 (https://github.com/ferdousbhai/ghost/releases/tag/v0.3.0); 0.2.3 in between. Minor rather than patch because this release removes the Claude Code chat runtime — the daemon now runs on one runtime (pi), and an installed home that still names the old one heals itself on read, so upgrading needs no owner action.

Two optdepends are gone with it (`claude-code` and `pnpm`, the latter only there for the optional Claude Agent SDK boundary); sources and digests refreshed as usual. Still happy to adjust anything the review needs.

ferdousbhai:
Updated to ghost 0.4.1 (https://github.com/ferdousbhai/ghost/releases/tag/v0.4.1); 0.4.0 in between, which is where the split into `ghost-runtime` and `ghost` landed. Sources and digests refreshed as usual.

One thing worth calling out in 0.4.1: the Chromium relay left for [its own repository](https://github.com/ferdousbhai/ghost-chromium-extension) and is installed from the browser, so it is no longer part of this package. `chromium` stays an optdepend because the daemon still drives a browser the owner already runs, and the relay endpoint is still what the packaged daemon serves.

I have also rewritten the description above to name no version. The branch is refreshed at every Ghost release while this is open, so a version in the description is stale the moment it is written; the branch head is the honest answer. Still happy to adjust anything the review needs.

ferdousbhai:
Updated to ghost 0.4.2 (https://github.com/ferdousbhai/ghost/releases/tag/v0.4.2); sources and digests refreshed, and they match the release's `SHA256SUMS`. A patch release: pi moves to 0.86.1, and the rest is release and test tooling. Nothing changes in the package's shape, depends, or install hooks.

PR #397, opened 2026-09-11
Package lazyjournal

bartex:
Packages [lazyjournal](https://github.com/Lifailon/lazyjournal), a TUI for journalctl, for the server edition's logs door.

Part of the server edition set: omacom/omarchy#11289 (runtime), omacom/omarchy-iso#175 (installer, builds it into the offline mirror). The menu degrades to `journalctl --pager-end` when it's absent, so this can land independently of the other two.


ryanrhughes:
Before this can build: `lazyjournal` has no `.omarchy/package.json`. Every package needs one.

bartex:
Thank you @ryanrhughes. I've added the `.omarchy/package.json` in d3a90f8

PR #405, opened 2026-09-11
Package rootless Docker and ONCE support

acrogenesis:
## Rootless ONCE package and service

Add a checksum-pinned ONCE v0.3.2 source build with the rootless proxy, Podman backup/restore and exec-completion fixes. The binary identifies as `v0.3.2-omarchy1`, provides `once-bin` for existing package/menu guards, and requires neither a Docker daemon nor the Podman Docker CLI shim. The vendor `once-bin` package remains separate; the preserved system unit is not automatically enabled. The source package is locally maintained so upstream synchronization cannot silently discard the patch.

Both settings variants conditionally install the source-provided `omarchy-once.service` user unit. Omarchy's optional installer enables it and user lingering after checking the selected rootless engine. Existing rootful ONCE workloads still require explicit migration.

Validation: real `makepkg` build/install in Lab, complete internal Go suite plus new regression tests, four actual settings payload builds, packaging/upstream-sync self-tests, and real Writebook content/credentials, backup/restore, replacement, scheduled tasks and reboot on both engines. Source package contents and patch are identical in the two companion proposals. Fresh ISO acceptance remains separate.

Rootless Docker needs Moby's rootless launcher and setup tool plus package-owned service defaults that cannot resolve daemon helpers through user-writable paths. This packages that runtime and supplies the companion Omarchy change with its rootless environment generator while keeping the separate Windows daemon root-only.

The `docker-rootless-extras` recipe now installs both `dockerd-rootless.sh` and `dockerd-rootless-setuptool.sh`. Its generated user service uses `PATH=/usr/bin` and `XDG_CONFIG_HOME=%h/.config`, preventing long-lived dockerd and RootlessKit helper resolution through mutable session paths. The Omarchy-specific changes live in `.omarchy/patches/rootless-service-boundary.patch`, so upstream refreshes reapply the boundary instead of silently dropping it; the downstream package release is `29.8.0-1.1`.

Both stable and development settings recipes conditionally install the rootless Docker environment generator and rootful socket drop-in when the selected Omarchy source contains them. That keeps builds against the currently pinned stable source compatible until it advances.

Companion Omarchy PR: https://github.com/omacom/omarchy/pull/11386

## Validation

- `makepkg --verifysource --skippgpcheck`
- isolated `makepkg -f --noconfirm` build of `docker-rootless-extras 29.8.0-1.1`
- package-worktree reproduction matches the patched source tree
- archive payload and generated `docker.service` inspection
- Bash syntax and whitespace checks
- clean merge-tree against current `master`
- real package installation and rootless runtime/migration validation in Omarchy Lab
- repeated independent Codex review; final round reported no valid actionable findings

## Screenshot

Rootless Docker running in Omarchy Lab with its user socket and a root-only Windows Docker socket

Writebook content restored through rootless Docker and ONCE in Omarchy Lab

Both settings variants normalize the installed rootless Docker helpers to root:root 0644 and their required parent directories to root:root 0755, including local `OMARCHY_SRC` builds that preserve source permissions. Six actual-recipe regression scenarios run under fakeroot in the existing self-test CI; migration trust checks remain strict.

Latest review: independent Claude Fable 5.1 and Codex followups report no actionable findings. Both build-isolation and self-tests pass in [CI run 34910006838](https://github.com/omacom/omarchy-pkgs/actions/runs/34910006838) after resolving the upstream workflow conflict.

### DIFF
--- a/pkgbuilds/docker-rootless-extras/.nvchecker.toml
+++ b/pkgbuilds/docker-rootless-extras/.nvchecker.toml
@@ -1,0 +1,7 @@
+[docker-rootless-extras]
+source = "github"
+github = "moby/moby"
+prefix = "docker-v"
+use_max_release = true
+include_regex = "docker-v.*"
+exclude_regex = ".*(rc|alpha|beta).*"
--- a/pkgbuilds/docker-rootless-extras/.omarchy/package.json
+++ b/pkgbuilds/docker-rootless-extras/.omarchy/package.json
@@ -1,0 +1,4 @@
+{
+  "source": "aur",
+  "upstream_commit": "5b48dc36fe09f9c04539b754c89e23d5898e8aeb"
+}
--- a/pkgbuilds/docker-rootless-extras/99-docker-rootless.conf
+++ b/pkgbuilds/docker-rootless-extras/99-docker-rootless.conf
@@ -1,0 +1,1 @@
+kernel.unprivileged_userns_clone=1
--- a/pkgbuilds/docker-rootless-extras/PKGBUILD
+++ b/pkgbuilds/docker-rootless-extras/PKGBUILD
@@ -1,0 +1,80 @@
+#!/bin/bash -e
+# Maintainer: Ľubomír 'the-k' Kučera <lubomir.kucera.jr at gmail.com>
+# Contributor: Hugo Osvaldo Barrera <hugo@barrera.io>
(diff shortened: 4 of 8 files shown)


ryanrhughes:
Superseded in part: docker-rootless and once are already on master. If the remaining omarchy-settings changes here are still wanted, please open a fresh PR against current master with just those.

PR #412, opened 2026-09-12
Publish what built instead of discarding the run over one package

omarchybot:
## The failure this prevents

`flea` 0.2.1 failed its test suite and stopped **every channel for a day**. `bin/build` exited 1, `bin/release` treated that as fatal and returned before sign/promote/sync — so the **twelve packages that built successfully in the same run were discarded**, rebuilt on the next tick, and discarded again. Nothing reached edge, rc or stable from 09-11 until the package itself was fixed.

## What changes

| file | change |
|---|---|
| `bin/build` | `exit 2` when at least one package built but others failed/blocked. `exit 1` still means a run that produced nothing to publish. Writes the failed names to a state file. |
| `helpers/paths.sh` | new `build_failures_file <mirror> <arch>` — build and release are separate processes, so the list must outlive the builder |
| `bin/release` | accepts `2`, runs the rest of the pipeline on what did build, reports the missing packages through the existing notifier, exits `2` |
| `bin/auto-release` | treats `2` as progress: clears the queue and the backoff |

`bin/build` already distinguished `failed` from `blocked` and never builds a package whose dependency failed. What it couldn't express was "some built" vs "nothing built" — both exited 1, so the caller had to assume the worst.

## The debatable bit

**A partial release clears the backoff instead of counting as a failure.** Counting it would reinstate the original bug on a longer timescale: four consecutive partials back a channel off to six-hour intervals, and one permanently broken package again decides when every other package ships.

The cost: a *transient* failure is no longer retried on the next tick — it waits for the next commit or version bump, which is also when it has a reason to succeed. Happy to flip this if you'd rather keep retry-on-transient; it's a one-line change.

## Safety

Nothing publishes that couldn't before. Blocked packages are still excluded from the build output, so no package ships without the dependency it was built against.

## Testing

`bash -n` clean on all four. Diffs reviewed against copies before applying.

**Note:** CI's `self-tests` job covers `sync-upstream`, `sync-rebuilds`, `omarchy-pkgs` and `omarchy-release` — **none of these four files**. A green check here does not mean this path was exercised. The partial path has not yet run against a real failure, since flea now builds.

### DIFF
--- a/bin/auto-release
+++ b/bin/auto-release
@@ -129,8 +129,19 @@ release_arch() {
   print_info "State file found: $state_file"
   print_info "Starting release workflow for $MIRROR ($arch)..."
 
-  if "$BUILD_ROOT/bin/repo" release --mirror "$MIRROR" --arch "$arch" --skip-prod-check; then
-    print_success "Release completed successfully for $MIRROR ($arch)"
+  # 2 is "published what built, reported what did not". That is progress, so it
+  # clears the queue and the backoff exactly like a clean run. Backing off for a
+  # package that is deterministically broken is what let one failure stall every
+  # other package: the release already reported it, and it retries on its own
+  # when a new commit or version changes the inputs that made it fail.
+  local release_status=0
+  "$BUILD_ROOT/bin/repo" release --mirror "$MIRROR" --arch "$arch" --skip-prod-check || release_status=$?
+  if ((release_status == 0 || release_status == 2)); then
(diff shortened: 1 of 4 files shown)


ryanrhughes:
Superseded: builds and publishing moved to CI in #511, which publishes per-package results by design; bin/release's all-or-nothing behaviour no longer gates anything.

Issue #378, opened 2026-09-10
OPR advisory sidecar (CVE metadata, refresh without rebuild)


calledtoconstruct:
## OPR advisory sidecar (CVE data that can move without a rebuild)

Arch already covers distro packages (`arch-audit` / Arch Security Tracker). OPR-built packages are silent there. Before we install or leave an OPR package in place, we should be able to answer: which known CVEs apply to **this published version**, and how fresh is that answer?

This is metadata OPR publishes, not a score maintainers put in the PKGBUILD.

## Venue

This issue is **OPR metadata only**.

In scope here:

- a published advisory sidecar for OPR package versions
- a scan/refresh pipeline that can update that sidecar without a new package upload
- who is allowed to write it (OPR-operated, not the PKGBUILD author)

Client audit, install preview, warn/deny policy, and omarchy-resolve behavior live in https://github.com/HxHippy/omarchy-resolve/issues/1. Pacman remains the authority on what is installed.

## Not the existing `trust` tier

omarchy-resolve already has `trust` on each route (https://github.com/HxHippy/omarchy-resolve):

| Field | Answers | Lives on |
|---|---|---|
| `trust` | Who owns it when this breaks (support boundary) | route |
| advisory / CVE fields (this RFC) | Known holes in this **version**, how fresh the scan is | OPR sidecar, keyed by package+version |

Do not reuse `trust`. Do not call the new field a score if it will be read as quality or as support.

Scanning and capability claims apply only to OPR `built-here` artifacts. A brew/flatpak/apt route is not something OPR scanned; the resolver must not print an OPR advisory band for it.

## Why not `.omarchy/package.json`

That file is build policy (`release_ring`, `pinned`, `source`, …) and is rebuilt with the package. CVE data has to refresh when the feed moves even if the bits did not. Maintainers must not be the writers.

## Proposal

Publish a **sidecar** next to the repo db, keyed by `(pkgname, pkgver, pkgrel, arch)` (or artifact id), independently signed and synced.

Fields for v1:

- `cve_ids`
- `cve_max_severity` (state the scale: CVSS or distro mapping)
- `advisory_as_of`, `scanned_at`, `scan_source`
- `scan_status`: `ok` | `stale` | `missing` | `error`

No composite `safety_score`. No capability tags (ports / fs / priv) in v1. If a UI needs a color, derive it client-side from `cve_max_severity` + `scan_status`, and label it “advisories”, not “safety”.

[...]

PR #416, opened 2026-09-12
feat: carry OPR advisory sidecar through channel advances

dyl-joseph:
## Why

Arch covers distro packages through arch-audit. OPR-built packages are silent there. This adds a CVE metadata sidecar beside the repo db, refreshed without rebuilds. It answers which known CVEs apply to each published OPR version and how fresh that answer is.

## Scope

- bin/sync-advisories plus helpers/advisory-helpers.sh ingest per-package feed JSON into omarchy.advisories.json.
- bin/repo gains an advisories subcommand forwarded to the repo host.
- bin/advance-channel merges sidecar entries forward with moved packages.
- docs/opr-advisory-sidecar.md, README entries, CI hook, tests/advisory-sidecar.sh.

## Tradeoffs

Merged sidecars drop their detached signature with a warning and stay fail-open until the next signed ingest. A known CVE is information, never a build failure. Missing, stale, and error states warn instead of blocking. Strictness lives with the client, not here.

## Blast Radius

Additive only. Existing commands keep their behavior. Advance gains a merge step that warns and continues on failure. Package archives are never written by the refresh.

## Verification

- bash -n syntax checks pass on all touched scripts.
- tests/advisory-sidecar.sh passes with exit 0. It asserts the sidecar validates, package archive hashes are unchanged across refresh, a new CVE updates only the sidecar, missing/stale/error never block, and banned fields never appear.
- Note: the suite needs GNU tar and GNU date. It fails on stock macOS for that reason. It passes on Linux and on macOS with GNU tools on PATH.

Refs #378.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,4 +44,5 @@ jobs:
               ./bin/sync-rebuilds --self-test
               ./bin/omarchy-pkgs self-test
               ./bin/omarchy-release self-test
+              ./tests/advisory-sidecar.sh
             '
--- a/README.md
+++ b/README.md
@@ -547,6 +547,7 @@ bin/package-worktree <package>       # Create upstream/patched/current scratch w
 bin/repo remove <package>            # Remove package
 bin/sync-upstream                    # Update packages that track a vendor release feed
 bin/sync-rebuilds                    # Bump pkgrel for packages whose dependencies moved
+bin/sync-advisories                  # Refresh the OPR advisory sidecar (CVE metadata, no rebuild; see docs/opr-advisory-sidecar.md)
 bin/clean-docker                     # Clear Docker images/cache (forces fresh rebuild)
 ```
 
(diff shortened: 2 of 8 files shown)


ryanrhughes:
Parking this rather than closing: since #511, channel advances and the release train are the last things still run from the host, and that surface is about to change again. Worth revisiting once the advance path is settled; the advisory sidecar idea itself is sound.

calledtoconstruct:
Published layout follow-up, from the note on #378: one signed file per artifact (`advisories/<pkgname>-<pkgver>-<pkgrel>-<arch>.json` plus `.sig`) instead of `omarchy.advisories.json`.

The feed files in this PR already have that shape. The channel document is what a refresh would rewrite for every package. This does not try to settle `advance-channel`; that stays on the host until the #511 cut-over. Copying the same advisory file when an artifact moves channel replaces the JSON merge.

Opening that change against this branch.

PR #422, opened 2026-09-13
fix: add Apple BCM43602 resume kernel

avillagran:
## Summary

- add a distinct `linux-bcm43602` kernel and matching headers package
- route the live-tested Apple BCM43602 (`106b:0133`) through brcmfmac's existing cold reprobe path after S3
- keep the stock `linux` package installed and bootable as a rollback
- leave every other BCM43602 device on the existing hot-resume path

## Root cause

On the affected Mac, the BCM43602 retains a non-zero PCIe interrupt mask after S3 even though its firmware message rings are dead. The existing register heuristic therefore selects hot resume and subsequent firmware commands time out. The patch bypasses that unreliable heuristic only for BCM43602 with Apple PCI subsystem `106b:0133` and uses the driver's existing serialized remove/reprobe recovery path.

## Live validation

Hardware:

- Apple MacBookPro12,1, x86_64
- Broadcom BCM43602 `14e4:43ba`
- Apple subsystem `106b:0133`
- `deep` S3 suspend

A/B results:

- stock `7.2.3-arch1-3`: reproduced the failure after S3 with 11 `brcmf_msgbuf_query_dcmd: Timeout` messages and unusable Wi-Fi
- patched kernel: four initial S3 cycles completed with one firmware reprobe per resume and no original failure signature
- final narrowed `106b:0133` module: rebuilt as an in-tree signed module, installed from the final package, booted with kernel taint `0`, and completed another real lid-close S3 cycle
- final cycle: `success=1`, `fail=0`, `failed_suspend=0`, `failed_resume=0`; firmware loaded again after resume; zero message timeouts, D3 failures, Oopses, or general protection faults
- post-resume: associated on 5 GHz and completed 5/5 Internet pings with 0% loss

No sleep hook, post-resume service, module unload/reload, or external reset script was present during validation.

## Package validation

- upstream kernel and Arch patch signatures verified with the listed official PGP keys
- source hashes and the packaged patch hash verified
- patch applies cleanly to Arch Linux `v7.2.3-arch1`
- `scripts/checkpatch.pl --strict --no-tree`: 0 errors, 0 warnings
- `makepkg --printsrcinfo` produces `linux-bcm43602` and `linux-bcm43602-headers`
- full package built natively on the affected Mac; the final narrowed brcmfmac module was rebuilt in-tree from the retained build tree and verified inside the installed package
- final installed module reports the expected kernel release, an in-tree marker, and the build-time kernel signature

## Maintenance note

This package follows the existing `linux-ptl` local-kernel pattern and therefore uses `skip_build: true`. Kernel version bumps require an explicit package rebuild plus renewed suspend/resume validation on the affected hardware.


ryanrhughes:
Held for review: this adds a full kernel package (`linux-bcm43602`, 13k lines) with `SKIP` on four source checksums. Kernel sources are signed, so `validpgpkeys` covers the tarball, but the patch and config should carry real sha256 sums. kwilczynski is requested since this is a kernel.

PR #426, opened 2026-09-13
Build omapresent for aarch64

scottjones:
The PKGBUILD deliberately withheld the aarch64 claim:

> `# Only x86_64 has ever been built and tested. aarch64 is plausible (the build`
> `# is plain qmake6 + make) but unproven, so it is not claimed here.`

It is proven now, so this claims it. No patches, no per-arch sources, no conditionals — the qmake6 + make build needed nothing.

Everything below was done on an Apple Silicon MacBook running Asahi Linux, natively on aarch64 rather than under emulation.

## It builds

`./bin/build --arch aarch64 --package omapresent`, twice from a clean worktree, producing an identical 1,861,744-byte package each time:

```
omapresent-0.1.3-1-aarch64.pkg.tar.zst
usr/bin/omapresent: ELF 64-bit LSB pie executable, ARM aarch64, stripped
```

`qt6-webengine` resolves on Arch Linux ARM (`extra` 6.11.2-1) and the binary links `libQt6WebEngineCore.so.6` and `libQt6WebEngineQuick.so.6`. That Chromium dependency was the most likely reason for this not to work.

The package is also complete rather than merely compiling: all nine hicolor icon sizes render through ImageMagick, and the desktop file, four vendored licence files, the agent skill with its reference docs, the theme-refresh hook and `welcome.md` are all present — 55 files, matching the x86_64 layout. Nothing in `package()` turned out to be arch-conditional.

## It runs

Run from the built package against the system Qt6 libraries. Every library resolves; nothing is missing.

`omapresent export --pdf` drives the whole QtWebEngine stack without a display, which makes it the most checkable evidence here:

```
PDF document, version 1.4, 3 page(s)
Page size:  960 x 540 pts
Creator:    Chromium
Producer:   Skia/PDF m140
```

Three slides from three `---` separators at the right 16:9 geometry, with Chromium and Skia named as the producer — so the render path executed on ARM rather than merely linking. The rendered output also shows the vendored libraries working: markdown-it parsed the deck, the theme colours applied, and KaTeX typeset `$E = mc^2$` as real math with a superscript.

The GUI works too. Under Hyprland it maps a genuine Wayland surface (`class='omapresent'`, `title='deck.md - Omapresent'`, 737x926) and shows the live split-pane editor — syntax-highlighted markdown on the left, WebEngine preview on the right, status bar with a live word count.

One note for anyone else testing on Asahi: both runs print `libva error: /usr/lib/dri/asahi_drv_video.so init failed`. That is Asahi's VA-API hardware video decode failing to initialise, unrelated to this package; Chromium falls back and renders correctly.

## Unrelated, mentioned rather than fixed here

On a fresh checkout the local `[omarchy]` pool under `pkgs.omarchy.org/$MIRROR/$ARCH` is empty, so `ttf-ia-writer` cannot resolve and the build fails — and `bin/build` still exits 0 in that case, reporting `Built: 0 / Failed: 1` alongside a success status. Anything consuming that exit code reads a total failure as green. Worth its own issue; I have kept it out of this diff.


ryanrhughes:
Superseded: omapresent already builds for aarch64 on master.

PR #438, opened 2026-09-14
Ship swap on zram for aarch64

cristim:
## Summary

- Install `default/systemd/zram-generator.conf.d/90-omarchy.conf` to `/usr/lib/systemd/zram-generator.conf.d/` on aarch64 too, in `omarchy-settings` and `omarchy-settings-dev`.
- Stop replacing `etc/sysctl.d/99-omarchy-sysctl.conf` with a network-only copy on aarch64, so aarch64 gets the same zram reclaim tuning as x86_64.
- Stop removing `etc/tmpfiles.d/omarchy-zswap.conf` on aarch64. It turns zswap off, which is what swap on zram needs, and aarch64 kernels enable zswap just like Arch's x86_64 kernel.
- Make the `zram-generator` optdepend architecture-neutral.
- systemd-oomd stays x86_64-only. `omarchy-settings` is repackaged from the same source, so its pkgrel goes to 2.

## Why

#275 grouped zram with the x86_64 boot and memory stack, but nothing in swap on zram is architecture-specific: it only configures memory and never touches the boot chain. Today aarch64 installs (Apple Silicon, DGX Spark, Snapdragon X) that have no disk swapfile run with no swap at all.

## Dependencies

This PR only ships configuration. Related work it relies on or that relies on it:

- **Needs `zram-generator` installed.** #172 adds it to the target package dependencies and omacom/omarchy#7553 installs it on existing systems. Without those, the drop-in stays inert on installs that lack the generator, exactly as on x86_64 today (omacom/omarchy#7522).
- **Apple Silicon fork:** maralcbr/omarchy-mx-mac#115 and maralcbr/omarchy-pkgs#92 make the same change for the Asahi builds. The fork's bundle updater rejects these paths until #115 ships, so the fork must release #115 before it takes this change.
- **Overlaps #380**, which rewrites the same header comment and the Limine/mkinitcpio lines just above the memory-stack lines in the aarch64 block. Whichever lands second needs a small rebase; I'm happy to do it.

## Validation

- Validated on an Apple M1 with 16 GB RAM (16K pages) on Asahi, kernel 7.1.13, with this configuration applied: `/dev/zram0` came up at 15.3G, zstd, swap priority 100, with the reclaim sysctls live. Under a 1 GiB memory cap, 3072 MiB of mixed data swapped 2183 MiB into zram (2.5G stored in 914M, about 2.7:1) and read back intact with no OOM events.
- Built `omarchy-settings` 4.0.3-2 natively for aarch64: the package contains `usr/lib/systemd/zram-generator.conf.d/90-omarchy.conf`, `etc/tmpfiles.d/omarchy-zswap.conf` and all eight `vm.*` lines, and no oomd drop-ins.
- Built `omarchy-settings-dev` 4.0.0.r2128.gb679363 (quattro tip) natively for aarch64 with the same result.

https://claude.ai/code/session_019NiURwGntmFfHfZf1rhY78


ryanrhughes:
Superseded: swap on zram for aarch64 is already on master in omarchy-settings.

PR #443, opened 2026-09-14
Add optional fm-cli package

ninepointlabs:
## Summary

Add `fm-cli` as an optional Omarchy package: Fastmail from the terminal. It is a small Go client for Fastmail's JMAP API with a full-screen mail app, scriptable commands that print JSON, and a live `watch` that follows the mailbox over Fastmail's push connection. It is the engine behind the [Fastmail plugin for Omarchy](https://github.com/ninepointlabs/omarchy-fastmail), which today installs it through `omarchy-mise-install`.

- Builds from the tagged source archive with `CGO_ENABLED=0`, `-buildmode=pie` and `-trimpath`, and runs the Go test suite in `check()`.
- Installs the `fm-cli` binary, README and MIT license. Runtime dependency is `glibc` only.
- Sign-in tokens go into the system keyring, so `gnome-keyring` (Secret Service) and `kwallet` are optional dependencies, as is `xdg-utils` for opening the OAuth consent page.
- Tracks releases with the `git_tags` upstream provider and the `{tag}` source archive, so `sync-upstream` picks up new tags and refreshes the checksum.
- On the fast ring, since the package is optional and self-contained.

The package is maintained by fm-cli's upstream author.

## Validation

- Verified the v0.3.0 source archive checksum.
- Clean `makepkg` build on x86_64, with `check()` passing; `fm-cli --version` reports `0.3.0`.
- `bin/sync-upstream fm-cli` reports no update against the checked-in version, and bumps a PKGBUILD rolled back to 0.2.4 to 0.3.0 with the correct checksum.

This is intended as an optional package and does not add the app to the default Omarchy installation.


ninepointlabs:
Friendly bump: Copilot’s Go pin is in `b7c2ca0` (`makedepends=('go>=2:1.26.0')`). Ready for review. If useful, happy to get `build-approved` on this PR or a line for `ninepointlabs` in `.github/VOUCHED.td` so future builds don’t sit skipped.

PR #447, opened 2026-09-15
Add Omawake and Omaspeak to edge

jacob-vincent-mink:
Adds Omawake 0.0.2 and Omaspeak 0.0.1 binary packages to edge for x86_64 and aarch64. Each includes its default CPU audio.cpp provider and accelerator setup documentation; users can configure an external runtime after installation.

The packaged user units remain disabled. Direct installation supports on-demand use. The separate guided Omarchy installers explicitly confirm service enablement, configure and validate the application, then start its user service: https://github.com/omacom/omarchy/pull/11883.

Validation:

- verified all four published archives against release checksums and GitHub asset digests
- built both x86_64 packages with makepkg and verified sources
- checked staged provider discovery, library symlinks, licenses, documentation, and passive service units
- regenerated .SRCINFO and verified edge-only metadata and build planning
- both application CI and both architecture release workflows passed; application line coverage is 90.21% for Omawake and 91.29% for Omaspeak

namcap was unavailable locally. Neither package installs or enables an accelerator runtime or starts a service in its install hooks.


spencerbull:
## Re-review at b836c23 (was 46306a12)

The delta since the last review is the move to the final releases (Omawake 0.0.2, Omaspeak 0.0.1), the aarch64 sources, the bzip2 drop, and the explicit provider-library install. All of that checks out: the four archive checksums match GitHub's release-asset digests, both architectures' archives contain every path `package()` reads, the binaries link only against glibc, gcc-libs and (for Omawake) alsa-lib, and neither binary needs bzip2 any more. Installed from the built x86_64 packages on a worker, `/usr/bin/omawake` and `/usr/bin/omaspeak` find `/usr/lib/<app>/libaudiocpp.so.0.1.0` with no environment override, both user units land in `/usr/lib/systemd/user` disabled and inactive, `setup check` runs, and removal leaves nothing behind.

Three small things were wrong, and I pushed a commit for each:

- **The four `.SRCINFO` and per-package `.gitignore` files are gone.** The repository's own `.gitignore` excludes both, and no other package tracks them. They were added under the retired `pkgbuilds/edge/` path, which that rule does not cover, and the move carried them along. The per-package `.gitignore` also negated the repository rule for its directory and its leading `*` hid every future file there: a patch or an `.omarchy/upstream.sh` dropped beside the PKGBUILD would never show up in `git status`.
- **`options=('!strip' '!debug')`.** The builder's makepkg.conf enables `debug` and `emptydirs`; with strip off, makepkg still takes the debug branch of `tidy_strip`, creates `usr/src/debug/<pkgbase>` inside the package, and ships it empty. Both packages built from the previous head carried that empty directory. Seven of the ten `-bin` packages here that disable strip already disable debug with it; the rebuilt packages contain no `usr/src`.
- **`license=('MIT' 'Apache-2.0' 'BSD-3-Clause')`.** Each package bundles libaudiocpp, which upstream's own notices describe as Apache-2.0 with BSD-3-Clause PocketFFT-derived code, and installs those texts under `/usr/share/licenses`. MIT alone described only the project's own source.

One thing is open and it is the maintainer's call, not a packaging defect: **`.omarchy/package.json` pins both packages to `channels: ["edge"]`**, which means `bin/repo advance` will never carry them into rc or stable (`package_moves_to_channel` refuses, and a `--mirror rc`/`--mirror stable` plan excludes them). The pin was added during the last review because the versions were release candidates; that reason has lapsed. Meanwhile omacom/omarchy#11883 adds menu entries that run `omarchy-pkg-add omawake-bin` for every user, and on the rc or stable channel that `pacman -S` ends in "target not found". Either the pin goes (delete the `channels` key and the packages flow edge → rc → stable like the rest) or the companion installers are gated to edge. Both pull requests are consistent only with one of those.

What ran, and where: everything executed ran on a disposable worker, never on the coordinator. `bin/build --mirror edge` built both packages for x86_64 and for aarch64 under QEMU from this head's tree, the repository self-tests passed in the CI container, and namcap reports only the arch-literal warning that per-architecture `source_*` arrays always trigger. Not tested: real microphones or speakers, OpenVINO or CUDA hardware, and running the aarch64 binaries. GPT-6 Astra at xhigh reviewed the branch in three rounds as an independent second opinion, before and after the fixes; it agreed on the tracked files and the channel pin (its independence from this session's reasoning is not guaranteed), and the license array was its finding. It also asked for ISC and Unicode-3.0 to be added for statically linked Rust crates; I left that out, because no package in this repository enumerates the licence graph of its vendored dependencies and the full cargo-about report is installed under `/usr/share/licenses`. A third round on the final head found nothing left to act on.

Waiting on: the maintainer, for the channel decision and the merge. The PR body's line about a regenerated `.SRCINFO` no longer applies.

🤖 Generated by Claude Fable 5.1 in Claude Code. Reviewed by GPT-6 Astra XHigh.

spencerbull:
## Channel decision: both packages go on the fast ring (c2f2345)

The maintainer has settled the open question from the last review. `.omarchy/package.json` in both packages now reads `{ "source": "local", "release_ring": "fast" }`, in place of `"channels": ["edge"]`.

Why the key was replaced rather than joined: `channels` is the outer bound on where a package may build, and `package_builds_for_mirror` checks it before it looks at the ring, so an edge-only list beside `release_ring` would still refuse the rc and stable builds the ring asks for. With the list gone and the ring set, the packages build natively for edge, rc and stable, each in its own image against its own base mirror, the same way the other prebuilt `-bin` applications here ship. That also resolves the conflict with omacom/omarchy#11883: `omarchy-pkg-add omawake-bin` no longer ends in "target not found" for a user on rc or stable once those channels have built.

What was checked: the repository's own `package_builds_for_mirror` answers edge only for the previous head and edge, rc and stable for this one, for both packages; `validate_package_metadata` accepts both files; and `bin/build --dry-run` plans both packages under `--mirror edge`, `--mirror rc` and `--mirror stable`. The x86_64 packages built from the previous head with `bin/build --package omawake-bin omaspeak-bin` install and run on a maintainer machine, and nothing about the PKGBUILDs changed since. No second model reviewed this commit; it is a two-line metadata change verified against the helpers that read it.

Two things this leaves for the author or the maintainer. The title and body still describe the packages as edge-only and should say fast ring. And neither package declares an `upstream` block, which 39 of the 47 fast-ring packages carry: without one `bin/sync-upstream` will not pick up new Omawake or Omaspeak releases, and with one a `min_release_age` is the usual companion for a binary that reaches stable directly. `schist-bin` is the closest model, a GitHub release with `digests: true` and one asset per architecture.

Waiting on: the maintainer, for the merge.

🤖 Generated by Claude Fable 5.1 in Claude Code.

PR #460, opened 2026-09-15
Add granola (AI meeting notepad)

shawnyeager:
Adds `granola` to the Omarchy package repo so `omarchy pkg add granola` works.

Granola only ships macOS/Windows. This recipe downloads the official `.dmg`, patches the Linux-rejected `platform=linux` API path, rebuilds Granola's `better-sqlite3-multiple-ciphers` fork against distro `electron44`, and installs `/usr/bin/granola` plus a desktop entry. The launcher icon is the lime squircle (`mac-icon.png`), not the transparent wordmark in `icon.png`.

Follows [tirtha4/Granola-for-Linux](https://github.com/tirtha4/Granola-for-Linux) (PR https://github.com/tirtha4/Granola-for-Linux/pull/8) and the existing AUR `granola` package (currently 7.559.2). This is 7.568.0, x86_64 only.

Built and installed locally as `granola 7.568.0-1` on Omarchy 4. Native-module smoke test (SQLCipher + `updateHook`) passes.

A companion PR to omacom/omarchy adds Install > AI > Granola once this package is in the `[omarchy]` repo.


ryanrhughes:
Before this can build: `pkgbuilds/granola/.omarchy/package.json` is present but the package has no readable `pkgver` when sourced with CARCH set, so the build planner cannot see it. Please check the PKGBUILD evaluates cleanly. Also note the source is a CloudFront URL rather than a vendor release page; a stable upstream URL or a GitHub release would be preferable.

shawnyeager:
> the package has no readable `pkgver` when sourced with CARCH set, so the build planner cannot see it.

Checked: with CARCH set, the PKGBUILD sources to pkgver=7.568.0 / pkgrel=1, and `bin/build --dry-run --package granola` queues it. What was missing was an upstream watch, so sync-upstream had nothing to read. Added one in 0b7527e against Granola's production versions API.

> a stable upstream URL or a GitHub release would be preferable.

Granola doesn't publish GitHub releases. The CloudFront URL is the official static CDN — `https://api.granola.ai/v1/download-latest` redirects to `https://dr2v7l5emb758.cloudfront.net/$pkgver/Granola-$pkgver-mac-universal.dmg`, and they tell package managers to pin that. A latest-redirect source would invalidate checksums on every release. Left source= on CloudFront; the watch uses https://api.granola.ai/v1/get-versions.

PR #466, opened 2026-09-16
Add brcmfmac-awdl-dkms for aarch64

brentkearney:
Patches for the BCM4387 wifi chip that ships in several Apple M1 and M2 computers that add support to it's firmware implementation of AWDL (Apple Wireless Direct Link) protocol. A DKMS module plus the root-side helpers that the [omdrop](https://github.com/brentkearney/omdrop-plugin) bar widget drives, so an Apple computer running Omarchy can receive files from an unmodified iPhone or Mac over AirDrop.

## Packaging

- `arch=('aarch64')`, so `package_supports_arch` skips it on x86_64 builds rather than failing them.
- One source: a 384 KB release tarball. The three kernel directories DKMS compiles are vendored inside it, pristine from `asahi-7.1.13-2` (`13aba96f`), and the AWDL patches are applied to a copy of them in `prepare()` — so the diff against upstream is exactly the eleven patches in the tarball, and the build runs offline in about two seconds. That matters here because aarch64 builds run under emulation.
- No kernel-header dependency, following #454 and the base-install guarantee.
- `patch(1)`, not `git apply`: the latter resolves paths against an enclosing repository's root, which silently applies nothing when the build happens inside a checkout.
- Upstream watch on the source repository's tags; the recipe is generated from the `PKGBUILD` there, so the two cannot drift.

## Failsafe

The module installs to `updates/dkms/`, which `depmod` prefers over the in-tree driver **without replacing it**. If a kernel upgrade breaks the out-of-tree build, stock `brcmfmac` loads and Wi-Fi keeps working — you lose AirDrop, never the network. That is the whole reason this is packaged as DKMS rather than a replacement module.

## Privilege surface

One root helper at `/usr/lib/omdrop/omdrop-discoverable`, reachable through a named polkit action (`allow_active=yes`, `auth_admin` for remote and inactive sessions). It needs `CAP_NET_ADMIN` for nl80211 vendor command passthrough and `CAP_NET_RAW` for the action frames AWDL synchronisation requires. The action binds to that one root-owned path. Nothing else in the package is privileged, and the file receiver runs as the user.

## Validation

- `makepkg` against the published tarball: checksum passes, eleven patches apply, packaged sources carry them (41 AWDL references in `core.c`).
- Those sources compile to `brcmfmac.ko` against `7.1.13-2-1-ARCH` headers.
- `validate_package_metadata` passes; `package_supports_arch` reports `aarch64` supported and `x86_64` skipped.
- Tested on an M1 Pro MacBook Pro (`j316s`/`t6000`) running Asahi. Receiving works end to end from iOS and macOS senders; sending is not implemented yet.

## Maintenance

The patches are pinned to one Asahi tag deliberately — a newer tree may need them rebased, and building against whatever is current would turn a rebase conflict into a runtime surprise. I maintain the patch set and will keep it current with Asahi tags; the re-vendoring sequence is documented in the source repository so it is not tribal knowledge.


brentkearney:
Pushed two corrections after re-reading how this repository carries packages.

**The patches are checked in now.** They were reaching the build inside the release tarball, which meant nothing here could diff, review or rebase them — the README is explicit that "ordinary source-code patches still belong beside PKGBUILD and are applied by `prepare()` as needed", and `yt6801-dkms` and `intel-ipu7-camera` are the shapes I should have followed. All eleven are listed in `source()` and checksummed individually; `prepare()` applies that copy. The tarball still carries the vendored kernel directories DKMS compiles.

**Dropped the generated-file header.** It pointed at an external source of truth, which contradicts "Omarchy owns every checked-in recipe". The recipe here is the recipe.

Rebuilt after both changes: every source validates, all eleven patches apply, and the packaged sources carry them (41 AWDL references in `core.c`).

On my `PUBLISHED_ARCHES` question in the description — I found the answer in the README ("packages whose `arch=()` lacks aarch64 are skipped, not failed"), so checked-in-and-unbuilt until aarch64 joins the scheduled pipeline is the expected state, not something I need solved.

PR #110, opened 2026-06-15
Fix: intl-ipu7-camera resume hook

spencerbull:
On an XPS 14 & 16 [Panther Lake] system running `linux-ptl 7.0.11.arch1-1`, the machine occasionally fails to wake from suspend or is laggy. 

Post issue, the journal shows the previous boot entered `s2idle` suspend and never logged a resume:

```text
2026-06-14T09:57:22-05:00 saber systemd-logind[1083]: Lid closed.
2026-06-14T09:57:22-05:00 saber systemd-logind[1083]: Suspending...
2026-06-14T09:57:23-05:00 saber systemd[1]: Stopping v4l2-relay daemon service for ipu7...
2026-06-14T09:57:23-05:00 saber systemd[1]: v4l2-relayd@ipu7.service: Deactivated successfully.
2026-06-14T09:57:23-05:00 saber systemd[1]: Stopped v4l2-relay daemon service for ipu7.
2026-06-14T09:57:23-05:00 saber systemd-sleep[41205]: Performing sleep operation 'suspend'...
2026-06-14T09:57:23-05:00 saber kernel: PM: suspend entry (s2idle)
```

The next entry is the forced reboot the following morning:

```text
2026-06-15T08:31:10-05:00 saber reboot: system boot 7.0.11-arch1-1-ptl
```

There is no matching `PM: suspend exit`, no `System returned from sleep`, and no post-resume userspace log for this failed suspend. That points to a kernel/firmware/device resume hang rather than just a compositor or display-session failure.

## Relevant Package Behavior

The installed `/usr/lib/systemd/system-sleep/camera-sleep-hook` is sourced from this package:

```text
pkgbuilds/intel-ipu7-camera/camera-sleep-hook
```

Current `post)` resume behavior unloads and reloads camera kernel modules automatically:

```bash
rmmod ov08x40 2>/dev/null || true
rmmod intel_ipu7_isys 2>/dev/null || true
sleep 1
modprobe intel_ipu7_isys 2>/dev/null
sleep 1
modprobe ov08x40 2>/dev/null
systemctl reset-failed v4l2-relayd@ipu7.service 2>/dev/null || true
systemctl start v4l2-relayd@ipu7.service 2>/dev/null || true
```

This logic was introduced by commit `07f5231 Fix the resume-from-sleep case`, which changed the older behavior from restarting `camera-init.service` to directly reprobeing `intel_ipu7_isys` and `ov08x40`.

## Evidence Implicating The Module Reprobe Path

[...]

PR #469, opened 2026-09-16
Update flea to 0.3.0 with its phone backends and shelf

thisisgm:
Flea 0.3.0 (https://github.com/thisisgm/flea/releases/tag/v0.3.0) adds gvfs-afc, gvfs-gphoto2, gvfs-mtp and usbmuxd for phones and cameras, ships its bar plugin under /usr/share/flea/shelf, and the copyfile.rs check follows the line the #110 fix changed; supersedes #340.


ryanrhughes:
Superseded: flea 0.3.0 with the gvfs backends is already on master via the upstream sync. (It currently fails its own test suite on build, which is a separate issue.)

Issue #467, opened 2026-09-16
linux-omarchy / linux-omarchy-bore: CONFIG_ANDROID_BINDER_IPC_RUST not set, breaks Waydroid (linux-ptl has it enabled)


isaackiii:
## What happened

After upgrading to v4.0.4, Waydroid stopped starting. `waydroid session start`
hangs and the container's journal fills with:

    [gbinder] ERROR: Can't open /dev/anbox-binder: No such process

Isolated the cause outside of Waydroid entirely — mounting a fresh `binder`
filesystem fails with "unknown filesystem type":

    $ unshare --user --map-root-user --mount --fork bash -c \
        'mount -t binder binder /tmp/bfs && ls /tmp/bfs'
    mount: unknown filesystem type 'binder'

    $ grep -i binder /proc/filesystems
    (nothing)

## Root cause

`linux-omarchy`'s config has Android binder support compiled out entirely:

    pkgbuilds/linux-omarchy/config.x86_64:
    # CONFIG_ANDROID_BINDER_IPC is not set
    # CONFIG_ANDROID_BINDER_IPC_RUST is not set

Same in `linux-omarchy-bore`. But `linux-ptl`, built in this same repo, has it
**enabled**:

    pkgbuilds/linux-ptl/config.x86_64:
    CONFIG_ANDROID_BINDER_IPC_RUST=y
    CONFIG_ANDROID_BINDER_DEVICES=""

Since all three configs were added in the same commit
(fbd4e34 / "Add new base and BORE-enabled Linux kernel packages for Omarchy"),
this looks like an oversight rather than an intentional decision to drop
Android/Waydroid support.

## Impact

Anyone running Waydroid loses it silently on upgrading to the new default
kernel — no warning, just a hang and a wall of crashed
`android.hardware.*`/`hwservicemanager`/`vndservicemanager` processes. The old
`linux` kernel package still works, but `linux-omarchy` is now the Limine
default (via migration 1789325478), so it's an easy trap to fall into without
knowing why.

## Ask

Enable `CONFIG_ANDROID_BINDER_IPC_RUST=y` (and `CONFIG_ANDROID_BINDER_DEVICES=""`)
in `linux-omarchy` and `linux-omarchy-bore`, matching `linux-ptl`.

## System

- Omarchy 4.0.4
- linux-omarchy 7.2.5-3
- MSI Stealth 15M A11UEKV, x86_64

kwilczynski:
[...]

PR #474, opened 2026-09-16
Add Android Binder support back to the base and BORE kernels

kwilczynski:


kwilczynski:
While at it, I also re-added missing signature files, as each patch has been signed.

kwilczynski:
Related to:

- https://github.com/omacom/omarchy-pkgs/issues/467

PR #480, opened 2026-09-16
Add HP EliteBook X G2i audio and camera support on linux-omarchy

meirdick:
Replaces #124, which targeted `linux-ptl`. Omarchy 4.0.4 boots `linux-omarchy`
and no longer installs `linux-ptl`, so the kernel side moves there.

On a stock install the HP EliteBook X G2i (Panther Lake, board 8E86, SKU
`D4DU1UT#ABA`; the X Flip G2i `D4VJ0UA#ABA` is the same board) has no
speakers, no microphone and no webcam. With this PR it has four-speaker stereo
with the factory calibration applied, a working microphone and headset jack,
a webcam on the hardware ISP, and a resume that the amplifiers no longer hold
up.

### Kernel

Eleven patches added to `pkgbuilds/linux-omarchy`, numbered into the existing
ranges with gaps kept. Every patch applies at offset 0 and fuzz 0 on 7.2.5-4,
was regenerated with `git format-patch` from a tree with the base applied, and
carries a full header (subject with subsystem prefix, body, Signed-off-by, and
the upstream SHA or list `Link:` where one exists). They are unsigned; the
`source=()` entries have no `.sig` companion, the way `7b11c97` ("Fix
unsigned patch sources for Omarchy kernels") listed them.

| Patch | Subject | Upstream status |
|---|---|---|
| `0520` | ASoC: Intel: soc-acpi-intel-ptl-match: add HP EliteBook X G2i | local only, stopgap: Intel does not take new PTL match entries (Bard Liao, 2026-06-29); HP BIOS Y90 99.15.23 reportedly lets the SDCA machine generator find the amps (LP#2164504), unverified here |
| `0521` | ASoC: tas2783: do not block system resume on peripheral re-attach | local; series prepared |
| `0522` | ASoC: tas2783: re-initialise amplifiers from .startup after system sleep | jml's v2 3/3, patchwork 14763920; stalled after Mark Brown's 2026-08-31 review ("feels like a bodge"); expected to be replaced by Baojun Xu's v4 (patchwork 14815978) |
| `0523` | soundwire: stream: append ports when a second DAI joins a stream | local; series prepared |
| `0524` | ASoC: Intel: sof_sdw: split a four-amp TAS2783A link into two PDIs | local; series prepared |
| `0525` | ASoC: tas2783: take one stream channel per mono amplifier | not sent; on 7.3 `173cb3147ba2` does the hw_params half in the codec and the `.set_tdm_slot` plumbing has no remaining purpose, so drop the whole patch there; still needed on 7.2.y |
| `0526` | ASoC: tas2783: keep the firmware image for re-initialisation | local only; re-evaluate against Baojun Xu's v4, which avoids the download differently |
| `0527` | ASoC: tas2783: read factory calibration from every vendor GUID | local; reply prepared to jml's v2 2/3 (patchwork 14763919), now carries both HP GUIDs (8E86 and 8EA1); upstream version sits on `3e56757584d6` |
| `0528` | ASoC: tas2783: do not fail runtime resume on a late peripheral | local; series prepared |
| `0543` | media: ipu-bridge: Add support additional link frequency | in media-next as `f9536a8065d9` for 7.4; `b883954d842f` (7.4) restructures the table, so the hunk needs a rebase before it is dropped |
| `0802` | platform/x86: int3472: 50 ms handshake delay for the ov05c10 | superseded in media-next by `844982d3b993` (200 ms, Cc stable); on a 7.3 base `edd0dd59aca0` changes `.hid` to `.hids[]` first |

Nine patches from #124 are already in this base and were dropped: `0020`,
`0028` (carried as `0400`/`0401`), `0029` (`0350`), `0030` (stable 7.2.5),
`0055`/`0056` (stable 7.2.5), `0057` (`0510`), `0058`/`0059` (`0511`).

`pkgrel` is `4.1` for the local build here so it sorts between the published
`4` and the next `5`; set it to whatever the next release is.

### Packages

`hp-elitebook-x-g2i-audio`, `hp-elitebook-x-g2i-camera`,
`hp-elitebook-x-g2i-camhal-runtime`, `intel-ipu7-drivers` and
`libcamera-ov05c10` come over from #124 unchanged, then the audio and camera
packages switch their kernel dependency to `linux-omarchy` and drop the stock
`linux-headers` dependency as #454 did for the other DKMS packages. Package
contents are unchanged; `makepkg --printsrcinfo` shows the dependency change.

### Verification

Verified on the reference board on `linux-ptl 7.2.3.arch1-5` with the same
patches before the rebase: four amplifiers each on their own channel, left and
right where the chassis puts them; factory calibration found for all four
unique ids; camera on `engine=camhal` from a cold start; audio back after
suspend without the amplifiers holding up resume. A second board (@cyberKrakn, X Flip G2i) confirmed the
camera and the stereo topology on `7.2.3.arch1-4`, and reported the late-amp
runtime-resume failure that `0528` fixes; the retest with `0528` is pending.

For this rebase: `makepkg -o` applies all 102 patches with no offset or fuzz;
the seven objects the patches touch compile with zero warnings at `W=1`;
`linux-omarchy 7.2.5-4.1` builds to completion in 43 minutes on 8 threads
(both packages produced; the only compiler warnings in the log are two
`-Wmaybe-uninitialized` in `tools/lib/bpf/libbpf.c`, outside anything this
PR touches). In the built modules: `snd-soc-tas2783-sdw.ko` has
`set_tdm_slot`, `fw_program`, the HP GUID bytes, the `factory calibration
applied` and `starting the stream without it` strings, and no
`__pm_runtime_suspend`/`__pm_runtime_set_status` references; `snd-soc-sof-sdw.ko`
has the `splitting into 2 PDIs` string; `intel_skl_int3472_discrete.ko` the
`OVTI05C1` entry; `snd-soc-acpi-intel-match.ko` the dummy topology name; and
`sdw_stream_add_master()` in `soundwire-bus.ko` is 192 instructions against
157 in the stock 7.2.5-4 module.

Not yet verified: a boot of `7.2.5-4.1-omarchy` on the board. `linux-omarchy`
turns `INTEL_IOMMU_DEFAULT_ON` on where `linux-ptl` had it off; the IPU7 and
Xe have run only in identity mode here, so the first boot should check
`/sys/kernel/iommu_groups/*/type` and, if the camera or display misbehave, try
`iommu=pt`.

### On the next base bump

- 7.2.6: `0510` will conflict on `tas2783-sdw.c` because 7.2.6 contains
  `08331ff6fffa` and `bfa756f3b8bf`, which `0510` also adds. Once `0510` is
  refreshed, `0521` and `0528` should still apply; they were regenerated against
  the `0510` versions of those hunks.
- Any base with `f9536a8065d9`: drop `0543`.
- Any base with `844982d3b993`: drop `0802`.
- A 7.3 base: drop `0525` (`173cb3147ba2` does the same in the codec). `0523`
  and `0524` are still needed for four amplifiers.
- `0522` when jml's v2 lands or Baojun Xu's "firmware download status check"
  (v4, patchwork 14815978) is merged; re-check `0526` against the same patch.
  `0527` if the GUID table or his second `#define` is merged.
- 7.4 folds IPU7 and IPU7.5 into the `intel_ipu6` driver (`2f4ffd14b6c0`,
  "media: ipu6: Enable support for IPU 7 and IPU 7.5", a 45-patch series).
  With `CONFIG_VIDEO_INTEL_IPU6_IPU7` on, the Panther Lake IPU7.5 binds to
  `intel_ipu6` by default and the in-tree `intel_ipu7` steps back (module
  options in both drivers move it either way). The `intel-ipu7-drivers` DKMS
  modules and the CamHAL runtime are built around the IPU7 driver, so that
  base bump needs the camera stack re-tested, not just the patches
  re-applied.

Not a bug, so nothing to carry: `tas2783-sdw` requests `8E86-2-9.bin` and
the like and `linux-firmware` ships `8E86-2-0x9.bin` symlinked to
`15E2-0-0x8.bin`. That alias is Arch's `make dedup` collapsing byte-identical
files, not an upstream rename; the upstream `8E86` files are regular files.
The audio package's filename symlinks stay as they are.

Addendum: `hp-elitebook-x-g2i-audio` 1.1.0-6 pins the per-amp analog gain (`Amp Volume`) in UCM. On this base the regcache restored 20 for the left pair and 12 for the right, an audible left-heavy imbalance; pinned, the tweeters measure within 0.5 dB.

Closes #124.


cyberKrakn:
@meirdick Rebase confirmed on the X Flip G2i (board `8E86`, `D4VJ0UA#ABA`). Built `pkgbuilds/linux-omarchy` from `hp-elitebook-x-g2i-omarchy` here (16 threads), installed it with its headers, `IgnorePkg` set, and DKMS rebuilt the camera modules against it.

`journalctl -k -b | grep -E 'factory calibration|not attached|splitting into 2 PDIs'`:

```
slave-tas2783 sdw:0:2:0102:0000:01:d: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:c: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:a: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:9: factory calibration applied (4 speakers in UEFI)
```

`not attached`: 0 lines.

`cat /sys/bus/soundwire/devices/sdw:0:2:0102:0000:01:*/status` -> `Attached` on all four (`9`, `a`, `c`, `d`).

`pactl list cards | grep 'Active Profile'` -> `Active Profile: HiFi` (5 sinks). Camera `engine=camhal` on the first open.

- **Suspend/resume**: s2idle 15:37 -> 16:10 (33 min). Calibration re-applied x4 on resume and audio came back without the amplifiers holding up resume — `0522` works.
- **IOMMU**: `cat /sys/kernel/iommu_groups/*/type | sort | uniq -c` -> `22 DMA-FQ`. DMA remapping is on rather than identity, and camera and display are both fine — no `iommu=pt` needed on this board.
- **`splitting into 2 PDIs` greps to 0, and that is the check line rather than the patch**: `0524` emits it with `dev_dbg`, so it never reaches a plain journal. The card still loads `sof-sdca-2amp-id2.tplg` and plays stereo, so nothing is wrong — but if you want that string visible on the next retest it needs `dev_info`, or `dyndbg="file sound/soc/intel/boards/sof_sdw.c +p"`.

Also installed `hp-elitebook-x-g2i-audio 1.1.0-6` for the per-amp analog gain pin.

Good to see the whole stack land on the new base — thanks again for the work.

meirdick:
@cyberKrakn Thanks, that is a full pass: enumeration, calibration, resume, camera and IOMMU all match what I see here.

One question: which BIOS are you on (`cat /sys/class/dmi/id/bios_version`)? HP's `01.02.02` (sp174220, via Windows Update) changes how the amps are initialised and silences the right pair on `4.1`. Root cause found and fixed today, `4.2` adds `0529` for it. If you are still on `01.01.18`, hold that for now, you are the only tester on the old path.

On the `dev_dbg`: keeping it. Upstream wants that line at debug level, and `dyndbg="file sound/soc/intel/boards/sof_sdw.c +p"` is the intended way to see it.

meirdick:
**4.2: fix for HP BIOS 01.02.02**

The new BIOS (sp174220, via Windows Update) describes the amps' SmartAmp function. The tas2783 driver then inits from BIOS data instead of its fixed sequence, and keeps the DSP posture the firmware image programs per side: left amps 1, right amps 4 (reads the second channel). `0525` gave each amp a one-channel port, so the right pair had nothing to play.

`0529` skips the per-amp channel mask when the peripheral has a SmartAmp SDCA function. The link mirrors the stereo port to all four amps and each picks its own side. Old BIOS keeps the `0523`-`0525` path.

Verified on 8E86 / 01.02.02: four amps play, no crosstalk, sides within 1 dB. Testers on 01.01.18: no change for you.

cyberKrakn:
@meirdick We're on `Y90 Ver. 01.01.18` (board `8E86`, `D4VJ0UA#ABA`), so the old path. Holding the BIOS update as you asked and staying on `4.1` — the pass above stands, and nothing here exercises `0529`.

Good to know the `dev_dbg` is deliberate; I'll use the `dyndbg` line when I need it rather than asking for `dev_info`.

ryanrhughes:
Before this can build: `hp-elitebook-x-g2i-camhal-runtime` has no `.omarchy/package.json`; every package needs one. Two other things to fix while you're in there: both DKMS PKGBUILDs use `SKIP` for every source checksum (12 and 7 entries), and the README shows `sudo pacman -U` of locally built files. Pinned sources with real checksums please. kwilczynski is requested for the kernel side.

meirdick:
All three fixed, and the branch is rebased on current `master`.

- `hp-elitebook-x-g2i-camhal-runtime` now has `.omarchy/package.json`.
- Every source in both DKMS packages has a real sha256 instead of `SKIP`, made with `updpkgsums`. The git sources were already pinned to a full commit SHA. pacman 7.1 checksums a git source from its checked-out tree, and I deleted the clone and re-fetched to confirm the hash comes back the same, so `makepkg --verifysource` passes on both.
- `TESTPLAN-v2.md` was the only doc with `pacman -U` of local builds. It was my own test plan, so it is out of the repo. `README.package.md` is now the only doc a user reads.

Also here: a fix to Intel's `ov05c10` driver in the camera package. It read a sensor register for VBLANK before checking runtime PM, so libcamera's control writes went to a powered-down sensor. Found through bprendie's Snapdragon port, which uses the same driver.

The kernel is rebased onto `linux-omarchy 7.2.5-6`, local pkgrel `6.2`. Our 12 patches apply and it builds.

cyberKrakn:
Boot confirmation for the 7.2.5-6.2 rebase on the X Flip G2i, same board as before: `8E86`, `D4VJ0UA#ABA`, stock BIOS `Y90 01.01.18` (holding the update as agreed).

Built `pkgbuilds/linux-omarchy` from the branch (16 threads, 106 patches applied with no offset or fuzz, ~44 minutes from source extraction to both packages), installed it with its headers (`IgnorePkg` set), DKMS rebuilt `ov05c10` 2.0.2, `ipu7-drivers`, `vision-drivers` and `v4l2loopback` against it, and booted `7.2.5-6.2-omarchy`.

`journalctl -k -b | grep -E 'factory calibration|not attached'`:

```
slave-tas2783 sdw:0:2:0102:0000:01:d: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:c: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:a: factory calibration applied (4 speakers in UEFI)
slave-tas2783 sdw:0:2:0102:0000:01:9: factory calibration applied (4 speakers in UEFI)
```

`not attached`: 0 lines. All four amps `Attached` in `/sys/bus/soundwire/devices/sdw:0:2:0102:0000:01:*/status`. `pactl list cards` -> `Active Profile: HiFi`, 5 sinks.

Camera: `hp-elitebook-x-g2i-camera` 2.0.2-3 with the ov05c10 VTS fix installed and DKMS-rebuilt. First open after boot starts `engine=camhal` (`sensor_starts=1`, `sensor_failures=0`, ~390 frames streamed), and it opens again cleanly after a resume (`sensor_starts=2`, `sensor_failures=0`, another ~300 frames).

Suspend/resume: s2idle entry/exit clean (12 s under `rtcwake`), factory calibration re-applied x4 on resume, all four amps `Attached`, `HiFi`, 5 sinks, and resume returned without the amplifiers holding it up.

IOMMU: `cat /sys/kernel/iommu_groups/*/type | sort | uniq -c` -> `22 DMA-FQ`; camera and display both fine, no `iommu=pt`.

Our path is BIOS `01.01.18`, so nothing here exercises `0529`; `0523`-`0525` is what booted.

Issue #479, opened 2026-09-16
linux-omarchy / linux-omarchy-bore: backport TTM swapout fix for amdgpu hard lockups after hibernate (drm/amd#5387)


ciryon:
## What happened

On AMD APUs, the machine hard-locks minutes to hours after resuming from hibernate. It can also happen during heavy GTT swapout without any hibernate, for example when loading large llama.cpp/Ollama models. A GPU client's command submission oopses while holding the TTM LRU spinlock. Every other GPU client then soft-locks, input stops working, and only a power reset helps.

I hit this twice on a Framework Laptop 13 (AMD Ryzen AI 300 Series), with Hyprland and suspend-then-hibernate. The second time, the kernel died 8 seconds after resume:

    PM: hibernation: hibernation exit
    BUG: kernel NULL pointer dereference, address: 0000000000000090
    RIP: 0010:ttm_lru_bulk_move_tail+0x158/0x1d0 [ttm]

Upstream report: https://gitlab.freedesktop.org/drm/amd/-/issues/5387 (mine is #5517, closed as a duplicate). It has been reported on Phoenix, Hawk Point, Strix and Strix Halo, on kernels 7.0 through 7.2.2.

## Root cause

Commit b2ed01e7ad3d ("drm/ttm: Fix ttm_bo_swapout() infinite LRU walk on swapout failure") moved the bulk_move cleanup in `ttm_bo_swapout_cb()` under `if (!ret)`. `ttm_tt_swapout()` returns the number of pages swapped out on success, so the cleanup never runs. Swapped-out resources stay in their bulk_move range. Once one of them is freed, `pos->first`/`pos->last` point at freed memory, and the next submission hits a use-after-free.

The `linux-omarchy` 7.2.5 source is still affected (`drivers/gpu/drm/ttm/ttm_bo.c`):

    ret = ttm_tt_swapout(bdev, tt, swapout_walk->gfp_flags);
    if (!ret) {

## Fix

Two commits are in `drm-misc-fixes`, both reviewed by Christian König and tagged `Cc: stable@vger.kernel.org # v7.1+`:

- `3db7d7d58341` "drm/ttm: fix swapped-out resources never leaving their bulk_move range"
- `fcfe64715b42` "drm/ttm: apply the swapout bulk_move fix to the intended condition"

[...]

PR #484, opened 2026-09-16
Add fix for swapped-out TTM resources never leaving their bulk_move range

kwilczynski:


kwilczynski:
Related:

- https://github.com/omacom/omarchy-pkgs/issues/479

PR #494, opened 2026-09-17
fix(linux-omarchy): drop missing 0220 patch signature requirement

jandrusk:
## Summary

`pkgbuilds/linux-omarchy/PKGBUILD` and `pkgbuilds/linux-omarchy-bore/PKGBUILD` list `0220-x86-amd-zen5-tlb-sizes.patch{,.sig}` in `source`, but only the `.patch` file exists in the tree. There is no `0220-x86-amd-zen5-tlb-sizes.patch.sig`, so signature-checked builds fail when makepkg tries to fetch/verify the missing `.sig`.

## Fix

- Drop `{,.sig}` for this file only, so `source` references `0220-x86-amd-zen5-tlb-sizes.patch` alone.
- Realign `b2sums` / `sha256sums` by removing the `SKIP` slot that belonged to the missing `.sig`.
- Keep the existing (correct) checksums for the patch itself; verified with `b2sum` / `sha256sum` on the tree copy:
  - b2: `e59c4da548b2ea4fd6144886e3e2e765b3e73972121e9e73dbc05c066834b2b35e6ef4c17617e77efc90260566bf7c0910d2138d8d5c1c3b6094eeac9de64cfe`
  - sha256: `718b40e15350049e03c0ed281b9257fca5432b8dec2e0e145e8b26507ee8892d`
- Same change applied to **linux-omarchy-bore**.

## Why not add a `.sig`?

The signature file was never present in the repository. Forging a maintainer GPG signature is not appropriate. Integrity of the patch remains covered by the checksum arrays. Maintainers can re-add a real `.sig` later if desired and restore `{,.sig}` plus a `SKIP` checksum slot.

No patch content changes and no `pkgver` bump.

## Test plan

- [ ] Confirm `0220-x86-amd-zen5-tlb-sizes.patch.sig` is absent upstream
- [ ] Confirm both PKGBUILDs: expanded `source` length matches `b2sums` / `sha256sums`
- [ ] `makepkg --verifysource` (or equivalent) for linux-omarchy / linux-omarchy-bore no longer fails on the missing `.sig`


kwilczynski:
@jandrusk thank you for sending this over, and sorry for troubles!

I added missing signature files. Hopefully, you should no longer have any issues with the build.

kwilczynski:
Superseded by:

- https://github.com/omacom/omarchy-pkgs/pull/487

PR #503, opened 2026-09-17
Package Omawake 0.0.3 and Omaspeak 0.0.3 with service-removal cleanup hooks

jacob-vincent-mink:
## Summary

   Adds two packaged local-AI applications and the removal-safety plumbing that
   keeps their user services from outliving the packages. Consolidates the two
   stacked review PRs (https://github.com/jacob-vincent-mink/omarchy-pkgs/pull/3 and https://github.com/jacob-vincent-mink/omarchy-pkgs/pull/4) into one
   reviewable change.

   ## What's in the packages

   ### omawake-bin 0.0.3 — local wake-word daemon
   - Pinned CPU provider library (audio.cpp C ABI, Moonshine Streaming Tiny +
     Silero VAD) — runtime-neutral, no vendor runtime baked in
   - CUDA/Vulkan/HIP served through externally supplied complete provider
     builds; OpenVINO GenAI for Whisper Base on Intel (sm_121a/GB10 verified)
   - `setup model --check-urls`: pinned-catalog URL health check (29 assets)
   - Spanish wake profile: multilingual Whisper Base INT8 OpenVINO row with
     `model.language` reaching inference and status
   - Moonshine Small/Medium evaluated on identical corpora and **deferred**
     (Tiny stays default; opt-in `asr_variant` profiles)
   - Connection-owned playback pauses (HoldPause protocol) so TTS playback
     cannot retrigger wake actions

   ### omaspeak-bin 0.0.3 — local-first text-to-speech
   - Kokoro 82M second family: Kokoro-capable packaged provider
     (supertonic;kokoro_tts + statically linked eSpeak-ng), 54 named voices +
     legacy numeric IDs, 24 kHz output
   - **eSpeak-ng data package is catalog-managed, not shipped in the core**:
     the Kokoro row pins it (18,383,744 B, sha256-verified release asset),
     installs it into the model directory, and the worker points
     `AUDIOCPP_ESPEAK_DATA` at it — the package root carries no model data
   - `setup model --check-urls` (18 pinned assets) + offline-import
     diagnostics with expected/observed values
   - Spanish speech profile (`model.language` reaches the native worker and
     daemon status); consistent `backend.requests` status shape
   - Supertonic streaming evaluated and **deferred** at the current pin

   ### Both packages — service-removal cleanup
   - `remove-user-services.hook`: PreTransaction ALPM hook (Remove-only,
     `AbortOnFail`) invoking a package-owned helper
   - `package-remove`: per-user cleanup that stops/disables units owned by
     the package binary (`/usr/bin/{app}`), removes generated user units and
     wants-links, runs as each owning user, covers offline users and
     lingering managers, and preserves configuration, models, masks and units
     pointing at development builds. Upgrades never trigger cleanup.

   ## Verification

   - Live install/service/uninstall/cleanup cycle for both packages on
     Arch x86_64: hooks stop/disable/remove package-owned units; units
     pointing elsewhere are preserved; config/models/caches untouched
   - 10-test removal regression suite (offline/online users, custom/masked
     units, effective overrides, failed stop, manager failure, root
     dispatch, both `package()` functions) — wired into CI
   - `makepkg --printsrcinfo`, shell syntax, and checksums verified
   - Upstream releases: [omawake v0.0.3](https://github.com/jacob-vincent-mink/omawake/releases/tag/v0.0.3) ·
     [omaspeak v0.0.2](https://github.com/jacob-vincent-mink/omaspeak/releases/tag/v0.0.2) →
     [v0.0.3](https://github.com/jacob-vincent-mink/omaspeak/releases/tag/v0.0.3)

   ## Notes

   - `pkgrel=2` on both: the removal hooks land in this same version, so a
     previously installed 0.0.3-1 without them still needs the hooks; the
     regression suite asserts the pkgrel contract
   - `.release-verification/` fixtures are intentionally not part of the PR
   - Older stacked PRs (https://github.com/jacob-vincent-mink/omarchy-pkgs/pull/3, https://github.com/jacob-vincent-mink/omarchy-pkgs/pull/4) on jacob-vincent-mink/omarchy-pkgs will be
     closed in favor of this consolidated PR


jacob-vincent-mink:
@spencerbull

spencerbull:
Reviewed `e0251112028a434a8c4f47fe0ef413b40dcefb2d`: **NEEDS AUTHOR**. Two cleanup defects affect both identical `package-remove` helpers:

1. **P2 — Preserve whitespace-formatted executable overrides** ([omawake:30](https://github.com/omacom/omarchy-pkgs/blob/e0251112028a434a8c4f47fe0ef413b40dcefb2d/pkgbuilds/omawake-bin/package-remove#L30), [omaspeak:30](https://github.com/omacom/omarchy-pkgs/blob/e0251112028a434a8c4f47fe0ef413b40dcefb2d/pkgbuilds/omaspeak-bin/package-remove#L30)). For a logged-out user, a drop-in containing `ExecStart =` followed by `ExecStart = /home/alice/build/omawake daemon` bypasses `grep '^ExecStart='`. The helper then removes the generated base unit and its enablement links despite the override selecting a development build. Systemd accepts whitespace around `=` ([parser](https://github.com/systemd/systemd/blob/v261/src/shared/conf-parser.c)). Recognize valid assignment whitespace and add preservation coverage for both helpers.

2. **P2 — Handle quoted configuration paths without evaluating shell syntax** ([omawake:20–21](https://github.com/omacom/omarchy-pkgs/blob/e0251112028a434a8c4f47fe0ef413b40dcefb2d/pkgbuilds/omawake-bin/package-remove#L20), [omaspeak:20–21](https://github.com/omacom/omarchy-pkgs/blob/e0251112028a434a8c4f47fe0ef413b40dcefb2d/pkgbuilds/omaspeak-bin/package-remove#L20)). `systemctl show-environment` emits a value such as `/home/alice/custom config` as `XDG_CONFIG_HOME=$'/home/alice/custom config'`. The `XDG_CONFIG_HOME=/*` case misses it and silently uses `~/.config`, leaving the generated unit in the actual configuration directory pointing at the removed binary. Systemd's [environment printer](https://github.com/systemd/systemd/blob/v261/src/systemctl/systemctl-set-environment.c#L56) uses shell quoting; the test stub emits raw values. Use structured output or safe decoding and cover a quoted path.

Verification: exact-head CI passed, including all 12 removal tests. Independently downloaded all four release archives and verified their SHA-256 hashes against the recipes and GitHub asset digests; checked packaged inputs, user units, ELF dependencies, and local helper/hook checksums. No additional concrete packaging defect found. The removal tests mock systemd. Disposable-worker provisioning failed because SSH to the configured hypervisor timed out, so I could not run real package transactions, the edge/rc/stable builds, or live service cleanup. Neither runtime reproduction nor a tested fix is claimed; no changes were pushed.

GPT-6 reviewed the change in Codex/T3 Code, with a separate GPT-6 xhigh reviewer in a named T3 child thread. Both reported the two findings, which were checked against the PR and systemd source; reviewer independence is not guaranteed because filesystem access is shared. Waiting on the author to address the findings and on worker availability for runtime verification. No merge performed.

🤖 Generated by GPT-6 in Codex/T3 Code. Reviewed by GPT-6 XHigh.

jacob-vincent-mink:
Both findings addressed in 5d50d5c. The two `package-remove` helpers stay byte-identical, so one change carries to both packages.

**1 — Preserve whitespace-formatted executable overrides**

Confirmed against the parser you linked: `parse_line()` strips the line and then both halves of the assignment (`strstrip(l)`, `strstrip(e)`), so `ExecStart = /home/alice/build/omawake daemon` overrides the executable just as surely as the unspaced form. The offline probe now matches an assignment the way the parser accepts one:

```sh
readonly execstart_assignment='^[[:space:]]*ExecStart[[:space:]]*='
...
if ! $online && grep -qs "$execstart_assignment" "$path.d/"*.conf; then
```

The gate that decides whether a unit file is the generated one (`owned_unit`) is left strict on purpose, so recognising the wider spelling cannot widen what gets deleted: only the exact generated shape is ever removed. Your reproduction against the installed script now reads

```
$ package-remove --user omawake /home/alice /run/user/1000      # no bus, logged out
rc=0 :: :: Preserving offline service with an executable override:
          /home/alice/.config/systemd/user/omawake.service
```
with both the base unit and the `graphical-session.target.wants` link intact.

**2 — Handle quoted configuration paths without evaluating shell syntax**

Also confirmed: `print_variable()` hands each value to `shell_maybe_quote(sep + 1, SHELL_ESCAPE_POSIX)`, which leaves an ordinary value alone and otherwise encloses it in `$'…'` with `\` and `'` the only two characters carrying a backslash. The helper now decodes that form character by character — nothing is ever handed to `eval` — and refuses a value that is neither a plain path nor a closed `$'…'` quote, rather than falling back quietly:

- `$'/home/alice/custom config'` resolves, and the generated unit and its enablement link are removed from the directory the manager really reads:
  ```
  rc=0 | unit gone: True | enablement link gone: True
  ```
- a quoted value cut across two lines (a control character in the path) aborts the cleanup with `:: Cannot tell which directory <home>'s manager reads <unit> from; removal aborted.` — unit, link and enablement preserved, and no `stop` issued at all;
- a relative `XDG_CONFIG_HOME` keeps the `~/.config` answer systemd itself uses there.

**Coverage**

`tests/oma-service-removal.py` goes from 12 tests to 15. The `systemctl` stub no longer emits raw values alone: the fixtures carry a mirror of `shell_maybe_quote(SHELL_ESCAPE_POSIX)` and hand over the very text a real manager prints. The new cases cover a quoted config home with a space, an apostrophe and a backslash, for both helpers; the unreadable quoted value; the relative one; and three whitespace spellings of an offline executable override, again for both helpers.

All 15 pass at this head, and the eight new sub-assertions fail against `e025111`'s helpers, so they do pin the two defects rather than merely accompany the fix. The decoder was additionally fuzzed against the mirror of the printer: 515 round-trip cases, no failures, and every value containing a control character refused instead of accepted.

**Recipes**

`pkgrel 3 → 4` with the helper's new checksum in both PKGBUILDs, and both `.SRCINFO` files regenerated with `makepkg --printsrcinfo`.

**CI at 5d50d5c**

`Tests / self-tests` and `Tests / build-isolation` both pass, the latter through real pacman transactions, which is what exercises the changed checksums and `pkgrel`. `Build changed packages` still cannot deliver a build here, for two reasons outside this commit: the `result` job holds it because the author is not yet in `.github/VOUCHED.td` (or labelled `build-approved`), and the `changes` job dies in its bootstrap, unable to check `bin/build-matrix` out of the PR head it fetched (`error: pathspec 'bin/build-matrix' did not match any file(s) known to git`). So live service cleanup, and the edge/rc/stable builds, remain as unverified for me as they were for you, pending the vouch and the base's build tooling.

spencerbull:
Re-reviewed the update from `e025111` through `5d50d5c`, then verified the fix pushed as [`380c392`](https://github.com/omacom/omarchy-pkgs/commit/380c3926f8ec7d2677fedc654876cf0f53106347). **Both previous findings are fixed. No remaining code findings; the hosted package-build gate is still blocked.**

The whitespace override and ordinary quoted-path regression tests pass on a disposable Omabot worker and fail against the old helpers. Real package removal preserves offline services with whitespace-formatted development overrides and removes generated units from configuration paths containing spaces, apostrophes, and backslashes.

The GPT-6 xhigh second review found one additional decoder defect: real `systemctl show-environment` C-escapes control characters, but the new parser rejected `\t`, `\n`, and octal escapes. A valid tab-containing `XDG_CONFIG_HOME` made both ALPM hooks abort uninstall before checking whether that account had an Oma service. I reproduced that with systemd 261.2 and an actual `pacman -R` transaction. Commit `380c392` decodes those escape forms without evaluating shell syntax, fixes the fixture's printer model, updates package checksums, and adds regression coverage. The new regression fails against `5d50d5c` and passes with the fix. The xhigh reviewer checked the final diff and found no further defects.

Verified on the disposable worker:

- All 16 removal tests, the remaining repository self-tests, and both generated `.SRCINFO` files pass.
- Both x86_64 packages build through `bin/build` against edge, rc, and stable: six builds, zero failures or skipped packages.
- Installation and `--version` work; `pacman -Qkk` reports zero altered files for both packages.
- Real ALPM removal succeeds with the fixed decoder, stops active test units, removes generated units from a quoted path containing a tab, and preserves offline custom overrides and settings. These were active `--help` oneshot units; model inference, physical audio, and native ARM execution were not tested.
- Fixture output matches the real manager printer for named controls, octal controls, quotes, backslashes, and trailing newlines. The worker was destroyed after evidence collection.

The separate [hosted package-build workflow](https://github.com/omacom/omarchy-pkgs/actions/runs/35408482148) remains red: its `changes` job cannot check out `bin/build-matrix` from the selected base or PR head. The vouch step never runs; the later “author is not vouched” message is a consequence of missing outputs, not an established author-trust failure. No hosted package build ran. That workflow/base-tooling mismatch still needs maintainer attention before this is merge-ready; the worker results above do not make a red gate green.

GPT-6 performed the review in Codex/T3 Code, with a named GPT-6 xhigh second-review thread. Its additional finding was independently reproduced on the disposable worker, and its final review found no further issues. Filesystem access is shared, so reviewer independence is not guaranteed. No merge performed.

🤖 Generated by GPT-6 in Codex/T3 Code. Reviewed by GPT-6 XHigh.

PR #506, opened 2026-09-18
Package Apple Silicon Wi-Fi defaults as a settings add-on

maralcbr:
Apple Silicon setup in omacom/omarchy#9835 needs a package-owned NetworkManager default. Add `omarchy-settings-asahi`, which owns `/usr/lib/NetworkManager/conf.d/wifi_backend.conf` with `wifi.backend=iwd`, replacing the runtime installer's generated configuration.

The add-on is aarch64-only, depends on `omarchy-settings`, `networkmanager`, and `iwd`, and is advertised as an optional aarch64 dependency by both settings variants. `omarchy-settings-dev` provides `omarchy-settings`. Selection belongs to the Apple image profile and existing Apple migration, so installing settings alone does not impose Apple Wi-Fi defaults on generic ARM or x86 systems. Administrator configuration under `/etc` retains precedence. Existing 4.0.4 source pins are preserved.

Validation at package commit `097cc7f0adee0adf4cef86494ac2bbfeb70667a5`:

- `makepkg` builds the add-on in an aarch64 container; source checksum and archive ownership/mode verified.
- Native M2 Max `makepkg` build passes with dependency checks enabled against the installed Omarchy settings provider, NetworkManager and iwd.
- The extracted package is read by NetworkManager's `--print-config` using isolated configuration directories: the vendor default selects iwd, and a same-name administrator override selects wpa_supplicant. No live network restart was performed.
- Package release version self-tests and 26 upstream-watch tests pass.

This is the package half of a preliminary cross-repository integration. Repository signing/publication and Apple image-profile inclusion must precede distribution of the dependent runtime. This PR does not enable ARM release channels or publish a runtime/image.


ryanrhughes:
Before this can build: `omarchy-settings-asahi` has `pkgver=1`, which will never sort against a future release; please give it a real version. The manifest is present. Note this PR also bumps omarchy and omarchy-settings to 4.0.4-2; those are the release pair and their pkgrel is set by the release process, so please drop those two bumps from this PR.

PR #509, opened 2026-09-18
linux-omarchy: fix poweroff/reboot hang with Thunderbolt/USB4 displays on 7.2

georgaberg:
## Problem

On Linux 7.2 every `poweroff` and `reboot` hangs forever on machines where a USB4/Thunderbolt device carries a DisplayPort tunnel: Thunderbolt monitors such as the Apple Studio Display, or docks with DP outputs. Userspace shutdown completes normally, the Plymouth logo stays on screen, and only the power button gets the machine off. Kernel 7.1.x and 6.18 LTS are unaffected.

Seen on a TianBei MACO mini PC (AMD Ryzen 7 255 / Hawk Point, AMD Pink Sardine USB4 host) with an Apple Studio Display. Journal boot history, measured from the last shutdown log line to the first line of the next boot:

| Kernel | Result |
|---|---|
| `linux 7.1.9` (20 boots) | 24–39 s, clean |
| `linux 7.2.3`, `linux-omarchy 7.2.5-3` (6 boots) | 59–226 s, every one needed a forced power-off |
| `linux-omarchy 7.2.5-7` built from this branch | 29 s, clean |

## Root cause

Commit f5cc545f5969 ("thunderbolt: Wait for tb_domain_release() to complete when driver is removed"), new in v7.2, makes `nhi_pci_remove()` block until the domain has been released. The asynchronous DP tunnel activation code (commit d6d458d42e1e) leaks a domain reference when a DPRX read is cancelled or still pending while the domain is stopped. Before 7.2 that only leaked memory; since 7.2 the completion never fires and `device_shutdown()` blocks forever.

The same bug is tracked at CachyOS/linux-cachyos#1047, which bisected it to the commit above and confirmed 7.2.4 and 7.2.5 affected.

## Change

Backport Sven Peter's six-patch fix series to `linux-omarchy` and `linux-omarchy-bore` as `0602`–`0607`, right after the existing `0600`/`0601` USB4 patches, and bump both recipes to `pkgrel=7`.

| File | Upstream commit |
|---|---|
| 0602-thunderbolt-hold-router-reference-per-hopid.patch | 032c59c7681c |
| 0603-thunderbolt-dp-tunnel-activation-callback-mandatory.patch | 12f5d8b85a66 |
| 0604-thunderbolt-fix-domain-ref-leak-on-dprx-cancel.patch | 1fd1f67c94d3 |
| 0605-thunderbolt-no-dp-tunnel-access-after-dprx-cancel.patch | 419fa32fa5bc |
| 0606-thunderbolt-mark-discovered-tunnels-active.patch | c222f80be55f |
| 0607-thunderbolt-tear-down-inactive-dp-tunnels-on-stop.patch | 0b457c6733f7 |

All six are in mainline through the [USB4/Thunderbolt fixes for v7.3-rc3](https://ratatoskr.run/linux-usb/2026/09/17537518/t) pull and are marked `Cc: stable` (`Fixes: d6d458d42e1e`), so they can be dropped from the recipe once a 7.2.y release carries them. 7.2.6 (2026-09-14) does not, and the 7.2 stable queue had no thunderbolt patches when I checked. The diffs are upstream's, with only the mail headers stripped to match the other patches in the recipe. The new entries have no `.sig`; I left signing to you.

## Validation

- **Hardware:** built `linux-omarchy 7.2.5-7` from this branch with `makepkg` (all PGP signatures and checksums verified, all 100 patches applied with `--fuzz=0`) and installed it on the affected machine. Poweroff and reboot now complete normally with the Studio Display attached. The `pkgbuild_sha256sum` in the tested package's `.BUILDINFO` matches the `PKGBUILD` in this PR.
- The six patches apply with small line offsets only and no fuzz. Stable 7.2.1–7.2.5 contain no changes under `drivers/thunderbolt`, so the base is identical to v7.2.
- `bash -n` and `makepkg --printsrcinfo` pass for both recipes; `source`, `b2sums` and `sha256sums` are aligned (196 entries for base, 198 for BORE).
- `linux-omarchy-bore` carries byte-identical patch files and the same `PKGBUILD` edit, but I did not build or boot the BORE variant.

## References

- CachyOS/linux-cachyos#1047: same bug, bisection, same patch set
- Stable submission of the series: https://ratatoskr.run/stable/2026/08/17477689/t


kwilczynski:
@georgaberg thank you for the report! Appreciated.

Let me have a look at the v7.3 pull request, to see what I can backport to our kernel. Assuming no major issues, then this will land in next release, so either v7.2.5-7 or v7.2.6-1, hopefully.

PR #517, opened 2026-09-18
Add Hype presentation editor package

dhh:
Adds Hype 0.1.0-4, a native Markdown presentation editor with visual slide ordering, Omarchy themes, inline video and animated WebP/GIF playback, and built-in PDF and rendered PowerPoint export.

The package installs the application, desktop launcher, icon, license, and user guide. Source is pinned to public `omacom/hype` commit `fc9b3ba`, with a SHA256 checksum generated by `makepkg --geninteg` using its Git archive convention. Qt image formats supplies WebP support. PowerPoint export is native and needs no Python dependencies. Animated WebP/GIF images are automatically converted to embedded H.264 MP4 with their layout, autoplay, and repeat settings preserved. Slides and converted animations are rendered at 4K (3840 × 2160), with automatic picture compression disabled. Fonts are rendered into slides and animation videos, so recipients need no matching fonts. libwebp supplies animation decoding with correct transparent frame disposal.

The manifest uses `source: local` to identify an OPR-maintained recipe. It does not describe the download protocol: [the metadata validator](https://github.com/omacom/omarchy-pkgs/blob/master/helpers/package-metadata.sh) accepts only `local` or `aur`, and [the upstream watch documentation](https://github.com/omacom/omarchy-pkgs/blob/master/docs/upstream-sources.md) uses `local` for recipes fetching upstream sources. The pinned Git URL belongs in PKGBUILD's source array.

Validation: built the x86_64 package with makepkg from the public repository, including dependency and SHA256 checks. OPR metadata validation passes. Hype's unit and GUI tests pass, along with all eight native PowerPoint export tests including LibreOffice. The Rails World 2026 trial exports all slides at 4K, including slide 13 as an embedded 18.8-second looping video. The 4K export was visually checked in Impress. Playback in Microsoft PowerPoint itself has not yet been verified. ARM64 is declared but was not built locally.


ryanrhughes:
Before this can build: `hype` pins a git commit with `sha256sums=('SKIP')`. A commit pin is fine, but please use `git+…#commit=…` with makepkg's git checksum convention rather than SKIP (see how other packages in this repo do it, e.g. the upstream watch docs). Also the manifest says `source: local` but the PKGBUILD fetches from git; please make those agree.

ryanrhughes:
Correction to my earlier comment: `source: local` is right for this package. It means the recipe is Omarchy-owned rather than synced from AUR, not where the PKGBUILD downloads from, so nothing to change there. Thanks for the real checksum.

The aarch64 failure is a builder bug, not this PR: the builder droplets register QEMU 7.2 (abandoned multiarch image), under which qmake's compiler probe fails on gcc 16. Fix is #518; once that is merged and deployed I'll rerun this job. The x86_64 build is already green.

PR #527, opened 2026-09-19
linux-omarchy: fix XPS 13 speaker-ID GPIO discovery

mranallo:
The Dell XPS 13 DX13260 loses built-in speaker playback when moving from
linux-ptl 7.1.8 to linux-omarchy 7.2.5-3. With this patch on the same 7.2.5-3
kernel, the SPI driver reads speaker ID 2 from two GPIOs, both amplifiers
load firmware and calibration, and the owner confirms audible playback
after reboot. The stock driver reads an incomplete ACPI speaker-ID
description and selects nonexistent spkid0 firmware.

Carry Richard Fitzgerald's unmodified [CS42L43 GPIO-reading patch](https://bugzilla.kernel.org/attachment.cgi?id=310816)
in linux-omarchy, with both source checksums and a package release bump.
The fix selects the existing firmware from the GPIO values; it does not
alias spkid0 to another tuning. The same patch was proposed for linux-ptl
in #419, which closed without merging.

Validation on an XPS 13 DX13260, SKU 0E54 / SSID 1028:0e54, BIOS 1.6.0:

- Owner-confirmed working audio on 7.1.8 and failure on 7.2.3 and 7.2.5-3.
- Stock 7.2.5-3 logs show FIRMWARE_MISSING for both amplifiers.
- Unmodified v7.2.5 driver rebuilt against installed headers has the same
  source version as the loaded stock module.
- Cirrus patch applies without fuzz and the affected module builds against
  7.2.5-3-omarchy headers.
- PKGBUILD syntax, source/checksum alignment, all local BLAKE2/SHA256
  checksums, and whitespace checks on the recipe and applied C source pass.
- Local DKMS build/install and boot-image rebuild pass. Module selection
  points at the override; the boot image contains no stale stock SPI driver.
- After reboot, patched driver version 0.1.0 is loaded on 7.2.5-3-omarchy
  (local DKMS test build of the same patch, plus a module-version marker).
- `spk-id-gpios = 2 from 2 GPIOs`; both amplifiers load firmware v4.5.9.
- Both log `Calibration applied` and both `CAL_SET_STATUS` controls are `Set`.
- Left tuning: PID 0x112481, SID 0x352202, TID 0x342200; right tuning:
  PID 0x112481, SID 0x352212, TID 0x342200.
- No `FIRMWARE_MISSING` in the post-reboot kernel log.
- Owner personally confirms audible built-in speaker playback after reboot.
- Full kernel build: not performed.

This is a candidate backport. [Upstream bug 221956](https://bugzilla.kernel.org/show_bug.cgi?id=221956)
remains under review, including confirmation of physical speaker type and
GPIO bit ordering. Existing reports of successful playback are not treated
as proof of that mapping on every machine. Maintainer/upstream review is
needed before shipping broadly.

Related: omacom/omarchy#9687, omacom/omarchy#11374.

Diagnostics, the local test package, and this packaging change were prepared
with Codex assistance. Playback was tested by the laptop owner; the log and
mixer results above were read from the machine after that reboot.


kwilczynski:
@mranallo thank you for the report!

This should have been included in our kernel, but it seems I have missed it.

I will add this patch back and ready new kernel release.

Sorry for troubles here!

mranallo:
@kwilczynski  No problem! Thanks for all your work.

PR #530, opened 2026-09-19
Package AMD SoundWire resume recovery service

tumbleweedlabs:
## Summary

- install the AMD SoundWire resume recovery user unit from Omarchy source
- keep release and development settings packages in sync

## Dependency

Companion to https://github.com/omacom/omarchy/pull/12491, which adds the service and recovery helper. This PR is intentionally limited to the two package install entries.


tumbleweedlabs:
Retracting this companion packaging change; keeping the fix local for further validation.

PR #533, opened 2026-09-19
Remove Hype Markdown MIME association

dhh:
Hype currently registers as a Markdown handler, which can make it the default application for ordinary `.md` documents. Remove that declaration from the installed launcher and bump the package release to 0.2.0-2.

The 0.2.0 source archive still contains the declaration, so the package strips it during installation. The development launcher and future source builds have also been corrected upstream in the working tree.

Validation: checked the generated desktop entry with `desktop-file-validate`; removing the local development association restores Omawrite as the Markdown default.


ryanrhughes:
No need to do this. Packages replace all owned files on update so just changing the file and building the package is sufficient. It'll replace the `.desktop` on update.

